PR #18, opened 2019-11-28
Add provider authentication

### LINKED ISSUE #17: Associate payment to a WM Provider
# Background

We've been thinking about the idea of looking at payments on a per-provider rather than per-user basis for the purposes of serving web-monetized exclusive content. Basically instead of making sure that we made enough money off of a given visitor, a site would check if they made enough money off of a visitor's WM provider before serving exclusive content. If a user runs their own provider then this system is equivalent to the current per-user accounting.

This has a few advantages. One is that a site can start serving WM content faster, without waiting for SPSP to intialize. The site can just see the user's provider and let them in right away once they validate the provider has paid. Another advantage is that this smooths out users with varying rates on a single provider.

# Proposed Changes

- Add a `Web-Monetization-Provider` header to SPSP queries. This includes a JWT or other signed authorization issued by the provider which can be verified by the site (the method of auth would be standardized so the site doesn't need to be aware provider-by-provider. public key would be hosted on well known URL). This lets wallets associate sessions to providers.

- Add a `"provider"` field or something similar (more generic) to the session API response. Then the site can account for sessions based on which provider they're from and pool them. If the wallet had a mechanism for automatically pooling sessions based on the provider it would also solve this problem and may be more convenient for the site.

- Add a way for the site to be notified of incoming funds. If the site is pooling by provider in its own systems then this means they don't need visitor to notify them every time they pay and then query the session and save it to their systems (which could be unreliable and requires unnecessary client-server communication). They could just get the wallet to tell them all the incoming funds in connection with their provider. This would have to be reasonably quick (~1s) so that the first visitor can access content and so that users running their own provider have a good experience. This is not necessary if a site can query by provider on the wallet directly.

CC @wilsonianb @adrianhopebailie @matdehaast

adrianhopebailie:
Fixes #17 

This PR proposes to allow a JWT to be provided with the session creation request that identifies the "issuer" (user wallet/WM provider).

The WM extension (and in future the browser) can request JWTs from the provider up-front that are already bound to new session ids that haven't been used to create a session yet. This avoids a roundtrip to the provider when starting a new session.

The JWT is not bound to a destination website but in future this could be done by including an "aud" claim. (Need to consider privacy implications although this could be done in bulk to avoid correlating users to sites they visit).

When the website does a spend against the session the wallet MAY apply that spend to the pooled funds that have been sent by all users from a single provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,35 +1,38 @@
-# Open Payments (OPay)
+# Open Payments
 
-Open Payments (OPay) builds on the OAuth and Open ID Connect frameworks to
-define standard authorization flows for payments using the Interledger protocol
-for a variety of payment use cases.
+Open Payments builds on the OAuth and Open ID Connect frameworks to define
+standard authorization flows for payments using the Interledger protocol for a
+variety of payment use cases.
 
 ## Background
 
-OPay is an evolution of the Simple Payment Setup Protocol (SPSP), providing
-support for new use cases and integrating into OAuth 2.0 so that existing
-identity infrastructure can be leveraged to handle payments authorization.
+Open Payments is an evolution of the Simple Payment Setup Protocol (SPSP),
+providing support for new use cases and integrating into OAuth 2.0 so that
+existing identity infrastructure can be leveraged to handle payments
+authorization.
 
-In the sections that follow, functions and features of OPay are compared with
-analogous functions and features of SPSP to help implementors that are upgrading
-to OPay from SPSP understand the differences (and similarities).
+In the sections that follow, functions and features of Open Payments are
+compared with analogous functions and features of SPSP to help implementors that
+are upgrading to Open Payments from SPSP understand the differences (and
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -511,10 +536,50 @@ If a `spend` fails because the amount spent exceeds the amount received the
 server MUST return a `402 Payment Required` response code and MUST NOT add to
 the spent amount.
 
+#### Authentication of the Sending System
+
+If the sending system/web monetization provider was identified during session
+creation, by providing a JWT as a Bearer token on the request, then the
+signature of the JWT MUST be validated AND the `jti` claim in the token MUST be
```
matdehaast: Confusing as before we say that the wallet MAY track the spending against a provider and then here we are using MUST. Seems conflicting or we should just say that this applies only IF the wallet chooses to support aggregation

PR #28, opened 2020-03-18
typo

austin-king:
AssetCode was 2, indicating $20/month not $2/month

### DIFF
--- a/documentation/docs/wallet-app.mdx
+++ b/documentation/docs/wallet-app.mdx
@@ -46,7 +46,7 @@ Content-Type: application/json
 {
   "amount": 200,
   "assetCode" : "USD",
-  "assetScale": 2,
+  "assetScale": 3,
   "interval": "P1M",
   "scope": "$issuer.wallet/alice"
 }
@@ -62,7 +62,7 @@ Content-Type: application/json
   "name": "//issuer.wallet/mandates/2fad69d0-7997-4543-8346-69b418c479a6",
   "amount": 200,
   "assetCode" : "USD",
-  "assetScale": 2,
+  "assetScale": 3,
   "interval": "P1M",
   "startAt": "2020-01-22T00:00:00Z",
   "scope": "$issuer.wallet/alice",


adrianhopebailie:
amount = 200
scale = 2
is $2.00

Am I missing something?

austin-king:
> Interledger amounts are integers, but most currencies are typically represented as fractional units, e.g. cents. This property defines how many Interledger units make up one regular units. For dollars, this would usually be set to 9, so that Interledger amounts are expressed in nanodollars.

^ From interledgerjs/ilp-connector

If it's going by orders of magnitude then that would mean that
(scale = 1, amount = 2, total = $2.00), (scale = 2, amount = 2, total = $0.20), (scale = 3, amount = 2, total = $0.02) therefore (scale = 3, amount = 200, total = $2.00)
right?
This stuff always trips me up haha

kincaidoneil:
> (scale = 1, amount = 2, total = $2.00)

I believe all implementations treat scale = 1, amount = 2 as $0.20, not $2.00. Here's the relevant code from [ilp-connector](https://github.com/interledgerjs/ilp-connector/blob/90dde497fc64ddefa69aef9ac3d3258e74547f9a/src/backends/ecb.ts#L141-L144); the scale is the number of orders of magnitude it shifts the decimal place.

From the settlement engine RFC:

> Asset amounts may be represented using any arbitrary denomination. For example, one U.S. dollar may be represented as \$1 or 100 cents, each of which is equivalent in value. Likewise, one Bitcoin may be represented as 1 BTC or 100,000,000 satoshis.
> 
> A standard unit is the typical unit of value for a particular asset, such as \$1 in the case of U.S. dollars, or 1 BTC in the case of Bitcoin.
> 
> A fractional unit represents some unit smaller than the standard unit, but with greater precision. Examples of fractional monetary units include one cent (\$0.01 USD), or 1 satoshi (0.00000001 BTC).
> 
> An asset scale is the difference in orders of magnitude between the standard unit and a corresponding fractional unit. More formally, the asset scale is a non-negative integer (0, 1, 2, …) such that one standard unit equals the value of 10^(scale) corresponding fractional units. If the fractional unit equals the standard unit, then the asset scale is 0.
> 
> For example, one cent represents an asset scale of 2 in the case of USD, whereas one satoshi represents an asset scale of 8 in the case of Bitcoin

austin-king:
> If the fractional unit equals the standard unit, then the asset scale is 0.

Got it, makes sense now. Thanks! closing.

PR #43, opened 2020-05-29
Add PayID discovery

adrianhopebailie:
A first attempt at adding support for PayID to Open Payments.

PayID is intentionally specified as a loose set of specifications (URI spec, Discovery spec and Protocol spec) to allow some mix and match. I have embraced that here to try and define a protocol that is as simple as Open Payments but achieves the same goals as the PayID protocol:

1. I have not used PayID Discovery but have gone for something simple that is almost the same as PayID Discovery manual mode but with a special path prefix `/$/` to avoid collisions and make it easy for Open Payments servers to write URL re-write rules for all requests of this type.
2. I have built on the existing Open Payments feature of doing a `GET` on account or invoice URLs with a different media type to get payment details. I added an XRP example which could be compared to the example in the PayID Protocol RFC to get an idea of the differences.
3. I considered the return types in PayID Protocol to be too complex. I didn't see the value of having different AddressDetails types. Instead, each payment details type can define the appropriate additional properties it will add to an account or invoice resource and a client (knowing the media type) will know how to parse those out and use them.
4. I couldn't see what value the payment network and environment properties added since these are implicit in the media type so have left them out.

I've preserved the guarantee that an account discovered from a Payment Pointer MUST support ILP but by using a PayID a client has more flexibility (but also a bit more work to do to discover the account it needs).

I'd like to still add something compatible with the Proof Of Control Signature but need to understand a little more about what problem this is solving.

To review locally:
```sh
git clone https://github.com/interledger/open-payments.git
cd open-payments
git checkout ahb/payid
cd documentation
yarn
yarn start
```

### DIFF
--- a/documentation/docs/discovery.mdx
+++ b/documentation/docs/discovery.mdx
@@ -111,7 +111,7 @@ payments** using Open Payments.
 
 Importantly, neither a PayID nor a Payment Pointer is sensitive in the same way
 that a payment card number is. They can be shared freely with no risk that funds
-will be pulled from the a discovered account without the account owner's consent
+will be pulled from the discovered account without the account owner's consent
 
 ## Account Metadata
 
--- a/documentation/docs/discovery.mdx
+++ b/documentation/docs/discovery.mdx
@@ -110,7 +110,7 @@ Users can share their PayID or Payment Pointers to both **get paid** and **make
 payments** using Open Payments.
 
 Importantly, neither a PayID nor a Payment Pointer is sensitive in the same way
-that a payment card number is. It can be shared freely with no risk that funds
+that a payment card number is. They can be shared freely with no risk that funds
 will be pulled from the a discovered account without the account owner's consent
 
 ## Account Metadata
--- a/documentation/docs/discovery.mdx
+++ b/documentation/docs/discovery.mdx
@@ -30,12 +30,11 @@ identifier called a **Payment Pointer**.
 
 ## Payment Pointers
 
-All entities in Open Payments are **resources** identified by, and addressable
-via, URLs in keeping with the
-[REST](https://tools.ietf.org/html/rfc7231#section-2) pattern. This includes
-[accounts](./accounts) which are the root resource in Open Payments. Discovery
-involves finding the URL of the account resource because all other resource URLs
-are relative to the account.
+In keeping with the [REST](https://tools.ietf.org/html/rfc7231#section-2)
+pattern, all entities in Open Payments are **resources** identified by, and
+addressable via, URLs. This includes [accounts](./accounts) which are the root
+resource in Open Payments. Discovery involves finding the URL of the account
+resource because all other resource URLs are relative to the account.
 
 > Example account URL:  
 > `https://wallet.example/accounts/dee69df8-be40-478f-9828-6267f4d42b01`
@@ -74,20 +73,47 @@ receiver).
 > Example Personal Payment Pointer:  
 > `$wallet.example/alice`
 
-Personal Payment Pointers are analogous to an email address or payment card
-number. Email addresses are given to counter-parties so that they can begin
-sending emails to the target inbox. Similarly, a payment card number can be
-given to a counter-party so that they can request payment from the target
+## PayID
+
+Another form of identifier that can be used by account owners is a PayID. A
+PayID identifies the account owner and not an individual account so it is
+possible for a client, given a PayID to discover multiple accounts they can
+transact with.
+
+These accounts may have different currencies and may have additional options
+making incoming payments.
+
+Discovery of accounts on an Open Payments server using a PayID is described in
+the [Open Payments PayID discovery protocol](./payid).
+
+### PayID vs Personal Payment Pointer
+
+While at first glance these may appear to be the same, they are intentionally
+different.
+
+A **Personal Payment Pointer** MUST point to an account that supports
+Interledger. A client that begins an Open Payments discovery algorithm with a
+Payment Pointer is guaranteed to discover an account that supports incoming
+payments via ILP.
+
+In contrast a PayID requires a more indirect discovery flow (via the account
+owner resource) it MAY be used to discover accounts connected to other payment
+networks or accounts denominated in different currencies.
+
+Both PayIDs and Personal Payment Pointers are analogous to an email address or
+payment card number. Email addresses are given to counter-parties so that they
+can begin sending emails to the target inbox. Similarly, a payment card number
+can be given to a counter-party so that they can request payment from the target
 account.
 
-Users can share their Personal Payment Pointers to both **get paid** and **make
+Users can share their PayID or Payment Pointers to both **get paid** and **make
 payments** using Open Payments.
 
-Importantly, a Payment Pointer is not sensitive in the same way that a payment
-card number is. It can be shared freely with no risk that funds will be pulled
-from the target account without the account owner's consent
+Importantly, neither a PayID nor a Payment Pointer is sensitive in the same way
+that a payment card number is. It can be shared freely with no risk that funds
+will be pulled from the a discovered account without the account owner's consent
 
-## Metadata
+## Account Metadata
 
 Account metadata is provided by an Open Payments Server as a JSON document that
 describes the various service endpoints that are available to:
--- a/documentation/docs/discovery.mdx
+++ b/documentation/docs/discovery.mdx
@@ -81,7 +81,7 @@ possible for a client, given a PayID to discover multiple accounts they can
 transact with.
 
 These accounts may have different currencies and may have additional options
-making incoming payments.
+for making incoming payments.
 
 Discovery of accounts on an Open Payments server using a PayID is described in
 the [Open Payments PayID discovery protocol](./payid).
--- a/documentation/docs/owners.mdx
+++ b/documentation/docs/owners.mdx
@@ -38,7 +38,7 @@ Create is not supported for owners.
 ### Get
 
 The following request is a non-normative example of a request for the meta-data
-for the account identified by the PayID: `alice$wallet.example`
+for the owner identified by the PayID: `alice$wallet.example`
 
 ```http
 GET /$/alice HTTP/1.1
--- a/documentation/docs/owners.mdx
+++ b/documentation/docs/owners.mdx
@@ -1,0 +1,114 @@
+---
+id: owners
+title: Owners
+---
+
+Owners represent the entity that owns an account. The owner resource can be used
+to discover the accounts owned by the entity, in order to setup a payment to or
+from the account.
+
+## Resource Representation
+
+The `application/json` representation of an owner resource provides the details
+of the owner as a JSON document.
+
+| Property | Type  | Description                      |
+| -------- | ----- | -------------------------------- |
+| id       | URL   | The URL identifying the owner    |
+| accounts | array | The accounts owned by the entity |
+
+The `accounts` property holds an array of account objects as described below:
+
+| Property   | Type   | Description                                                                       |
+| ---------- | ------ | --------------------------------------------------------------------------------- |
+| id         | URL    | The URL of the account resource                                                   |
+| assetCode  | string | The asset code of the amount (inherited from the underlying account)              |
+| assetScale | int32  | The scale of the amount (inherited from the underlying account)                   |
+| mediaTypes | array  | An array of media-types that can be requested when accessing the account resource |
+
+## APIs
+
+The owners API endpoint/resource URL is the URL discovered by the client through
+the [Open Payments PayID discovery](./payid) process.
+
+### Create
+
+Create is not supported for owners.
+
+### Get
+
+The following request is a non-normative example of a request for the meta-data
+for the account identified by the PayID: `alice$wallet.example`
+
+```http
+GET /$/alice HTTP/1.1
+Host: wallet.example
+Accept: application/json
+```
+
+The returned JSON document MUST contain the following claims:
+
+- `id` URL identifying the account owner.
+- `accounts` an array of account object as defined below.
+
+#### Accounts
+
+Each account object returned in the `accounts` array MUST contain the following
+claims:
+
+- `id` URL identifying the account. This is the account resource URL and can be
+  used as the basis for further Open Payments interaction on the account.
+- `assetCode` the currency of the account.
+- `assetScale` the natural scale of the account relative to the natural scale of
+  the asset. This is expressed as an integer which is the power of 10 that the
+  account balance must be divided by to get to the natural scale of the asset.
+  For example, if the asset is USD then a scale of 2 implies that the balance is
+  tracked in USD cents. (Must be divided by 100 to be show USD)
+- `mediaTypes` an array of strings that are valid media types for this account.
+
+#### Media Types
+
+By default, [account](./accounts) and [invoice](./invoices) resources are
+represented using the `application/json` media type.
+
+However, clients may request alternative representations which include
+additional payment details for payments into the account or against the invoice
+as described in the [payment details API](./payments).
+
+An example of such a media type is `application/ilp-stream+json` which will
+supplement the default representation with two additional properties the
+`ilpAddress` and `sharedSecret` which are used by the client to make payments to
+the account or against the invoice.
+
+The `mediaTypes` property of `account` objects returned in a response provide
+the set of media types that MAY be requested for the account.
+
+The list MUST include `application/json`.
+
+If the account is discoverable via a
+[Payment Pointer](./terminology#payment-pointer) the list MUST include
+`application/ilp-stream+json`.
+
+An error response uses the applicable HTTP status code value.
+
+The following is a non-normative example response:
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+  "id": "https://wallet.example/$/alice",
+  "accounts" : [
+    {
+      "id": "",
+      "assetCode": "USD",
+      "assetScale" 2,
+      "mediaTypes": [
+        "application/json", 
+        "application/ilp-stream+json" 
+      ]
+    }
+  ]
+}
+```
--- a/documentation/docs/payid.mdx
+++ b/documentation/docs/payid.mdx
@@ -15,7 +15,7 @@ for Interledger-enabled accounts.
 Said differently, any account identified by a Payment Pointer can be paid into
 using the [STREAM](terminology#STREAM) protocol.
 
-This feature allows clients setup payments, for uses cases such
+This feature allows clients to setup payments, for uses cases such
 [web monetization](./web-monetization), with a single API call, to get the
 payment details, and be guaranteed that the receiving account supports ILP as
 payment method.
--- a/documentation/docs/payid.mdx
+++ b/documentation/docs/payid.mdx
@@ -1,0 +1,74 @@
+---
+id: payid
+title: PayID
+---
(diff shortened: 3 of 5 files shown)


documentation/docs/discovery.mdx
```diff
@@ -74,20 +73,47 @@ receiver).
...
-given to a counter-party so that they can request payment from the target
+## PayID
+
+Another form of identifier that can be used by account owners is a PayID. A
+PayID identifies the account owner and not an individual account so it is
+possible for a client, given a PayID to discover multiple accounts they can
+transact with.
+
+These accounts may have different currencies and may have additional options
+making incoming payments.
```
adrianhopebailie, line 84:
```suggestion
for making incoming payments.
```

documentation/docs/discovery.mdx
```diff
@@ -74,20 +73,47 @@ receiver).
...
 
-Users can share their Personal Payment Pointers to both **get paid** and **make
+Users can share their PayID or Payment Pointers to both **get paid** and **make
 payments** using Open Payments.
 
-Importantly, a Payment Pointer is not sensitive in the same way that a payment
-card number is. It can be shared freely with no risk that funds will be pulled
-from the target account without the account owner's consent
+Importantly, neither a PayID nor a Payment Pointer is sensitive in the same way
+that a payment card number is. It can be shared freely with no risk that funds
```
adrianhopebailie, line 113:
```suggestion
that a payment card number is. They can be shared freely with no risk that funds
```

documentation/docs/owners.mdx
```diff
@@ -0,0 +1,114 @@
...
+the [Open Payments PayID discovery](./payid) process.
+
+### Create
+
+Create is not supported for owners.
+
+### Get
+
+The following request is a non-normative example of a request for the meta-data
+for the account identified by the PayID: `alice$wallet.example`
```
adrianhopebailie, line 41:
```suggestion
for the owner identified by the PayID: `alice$wallet.example`
```

documentation/docs/discovery.mdx
```diff
@@ -74,20 +73,47 @@ receiver).
...
-Users can share their Personal Payment Pointers to both **get paid** and **make
+Users can share their PayID or Payment Pointers to both **get paid** and **make
 payments** using Open Payments.
 
-Importantly, a Payment Pointer is not sensitive in the same way that a payment
-card number is. It can be shared freely with no risk that funds will be pulled
-from the target account without the account owner's consent
+Importantly, neither a PayID nor a Payment Pointer is sensitive in the same way
+that a payment card number is. They can be shared freely with no risk that funds
+will be pulled from the a discovered account without the account owner's consent
```
sublimator, line 114: will be pulled from **the a** discovered
typo ^

adrianhopebailie:
```suggestion
will be pulled from the discovered account without the account owner's consent
```

documentation/docs/payments.mdx
```diff
@@ -66,3 +66,43 @@ Content-Type: application/spsp4+json
...
+  "id": "https://wallet.example/alice",
+  "accountServicer": "https://wallet.example",
+  "assetCode": "USD",
+  "assetScale" 2,
+  "auth" : {
+    "authorization_endpoint": "https://auth.wallet.example/authorize",
+    "token_endpoint": "https://auth.wallet.example/token",
+  }
+  "xrplAddress": "XTVQWr6BhgBLW2jbFyqqufgq8T9eN7KresB684ZSHKQ3oDth",
+  "destinationTag": "543237"
```
aanchal4, line 106: PayID protocol supports `"application/payid+json"` media type which requires PayID server to return an array of _all_ payment accounts corresponding to a PayID. In such a case, the payment network and environment properties are not implicit in the media type. And so the JSON response MUST have payment network and environment fields as described in the PayID protocol RFC.

```
"addresses" :
   [
     {
       "paymentNetwork" : "xrpl",
       "environment" : "testnet",
       "addressDetailsType" : "CryptoAddressDetails",
       "addressDetails" : {
                    "address" : "XTVQWr6BhgBLW2jbFyqqufgq8T9eN7KresB684ZSHKQ3oDth"
            }
      }
    ]
```
Also an X-Address does not have a destination tag.

documentation/docs/payments.mdx
```diff
@@ -66,3 +66,43 @@ Content-Type: application/spsp4+json
...
+| xrplAddress    | XRP Ledger Address | The XRP Ledger address to use when making payments to the account or against the invoice |
+| destinationTag | string             | The destination tag to use when making payments to the account or against the invoice    |
+
+The following is a non-normative example where the client gets the details to
+pay an account via the XRP Ledger:
+
+```http
+GET alice HTTP/1.1
+Host: wallet.example
+Accept: application/xrpl+json
```
aanchal4, line 87: `application/xrpl+json` is not one of the standard/registered media-types. Payment networks could have different environments such as mainnet, testnet, devnet, etc. It is important to indicate the same in the client's `'Accept' `header as well as indicated in the response in some form (see the comment below for reasons). I would suggest to refer to PayID protocol RFC for more details on media-types.

`application/xrpl-testnet+json`

documentation/docs/payid.mdx
```diff
@@ -0,0 +1,74 @@
...
+1. Given a PayID conforming to the prescribed format, `acctpart "$" host`,
+   extract the values **acctpart** and **host**.
+2. Let **path** be the string `/$/` concatenated with **acctpart**
+3. Let **account owner URL** be a new URL as defined in
+   [WHATWG URL](https://url.spec.whatwg.org/#concept-url) where:
+   1. The URL's scheme is `https`
+   2. The URL's host is **host**
+   3. The URL's path is **path**
+
+For example, the PayID `alice$wallet.example` resolves to the URL
```
sappenin, line 57: Let's debate/discuss this a bit more. While I do like the simplicity, I don't like stepping on an operator's path-space.

So, it would be more ideal if we had a very simple well-known location for openpayments with a simple template that would allow operators to define this path.

I have not been persuaded that the Webfinger approach is "bad" or "too hard" or whatever -- in fact, I have a very simple template syntax, and supporting webfinger can be done with a single static file in the simplest of use-cases, all while providing full domain-customization.

In fact, I would argue that supporting a 2-line JRD in a static file is about as much work from a programming perspective as supporting this manual assembly. So, why not just use the solution that gives domain operators more power?

documentation/docs/payid.mdx
```diff
@@ -0,0 +1,74 @@
...
+resource and all APIs are relative to the account URL.
+
+The account URL is [discovered](./discovery) through a
+[Payment Pointer](./terminology#payment-pointer) which is an account identifer
+for Interledger-enabled accounts.
+
+Said differently, any account identified by a Payment Pointer can be paid into
+using the [STREAM](terminology#STREAM) protocol.
+
+This feature allows clients setup payments, for uses cases such
```
sappenin, line 18:
```suggestion
This feature allows clients to setup payments, for uses cases such
```

documentation/docs/payid.mdx
sappenin: One more thing - I think it will become necessary to support arbitrary meta-data for a PayID, which Webfinger supports very nicely (just add another 2-line JRD to an array).

For example, we'll likely want to define with a bit more fidelity some kind of "Wallet Discovery" protocol that allows software using OpenPayments to redirect a user to a wallet-link in order to make a payment.

Another example is any yet-to-be defined protocols that might need to facilitate PayID-level metadata.

adrianhopebailie: > I don't like stepping on an operator's path-space.

Why? What are the odds of a path collision?
If there is a collision it's simple to get around by simply using a subdomain.

You are putting technical purity at the top of the order of constituents and it should be at the bottom

adrianhopebailie:
@aanchal4 said:

> This is the cryptographic proof (an additional assurance for the PayID client) of the ownership of the on-ledger address.

Is there an equivalent for non-crypto payment addresses? I'm loathe to put something into Open Payments that makes it dependent on crypto.

documentation/docs/payid.mdx
adrianhopebailie: >  I think it will become necessary to support arbitrary meta-data for a PayID

In Open Payments that is done after discovery. Once the URL of the account owner is known we can add whatever additional meta-data we need to the owner resource.

documentation/docs/payments.mdx
adrianhopebailie: > application/xrpl+json is not one of the standard/registered media-types

No, but this is a different protocol. The media type we use for ILP is also different to what is proposed by PayID.

> Payment networks could have different environments such as mainnet, testnet, devnet, etc. It is important to indicate the same in the client's 'Accept' header as well as indicated in the response in some form (see the comment below for reasons).

My assertion is that this is an unnecessary over complication so for Open Payments I'm proposing that we exclude this.

A simpler way to do this is to only issue identifiers in test domains if they are for test networks: e.g. `alice$testnet.rafiki.money`

Or, if absolutely necessary, leave the live network unchanged and add a suffix to the media-type for the test network to disambiguate them.

> I would suggest to refer to PayID protocol RFC for more details on media-types.

I have reviewed the RFC but this is a different protocol. 

Open Payments is only dependent on the PayID URI RFC. This PR proposes a new discovery protocols and uses Open Payments for setup (as opposed to PayID Protocol).

documentation/docs/payid.mdx
sappenin: > Why? What are the odds of a path collision?

Path-collision isn't the only concern. What about operators that don't have fine-grained control over their domain-space (e.g., it's possible that just "adding a sub-domain" may not not work for certain operators).

sappenin: > You are putting technical purity at the top of the order of constituents and it should be at the bottom

This feels pejorative -- the comment doesn't actually make a technical case for why "robust discovery" should be at the bottom of the order as opposed to the top.

In my view, we're building a decentralized protocol here with open-payments. We need to be thoughtful about the restrictions/assumptions we make about it because if it ever becomes widely deployed, we won't be able to easily take these decisions back.

Overall, whether we use PayID Discovery or not here, I do feel like the `./well-known` space is the more appropriate location for open-payments meta-data and discovery. Perhaps we go back to the original OpenPayments Discovery endpoints? That location could still be easily derived from a PayID (whether using PayID Discovery or not) and we'll have a lot more flexibility in terms of what we put into `https://{host}/.well-known/open-payments`.

What do you think about that?

documentation/docs/payments.mdx
sappenin: > I have reviewed the RFC but this is a different protocol.

Yeah -- @aanchal4  and @adrianhopebailie, I think this is the disconnect between you. This proposal is re-defining some of what PayID can do by pulling it into the OpenPayments layer (specifically: where does one get the crypto address for a PayID). 

FWIW @adrianhopebailie (see my other comment above around where clients get address info), we've been doing a lot of thinking around this and have concluded that it makes sense to let the PayID server treat sometihng like an XRPL address on-par with a Payment Pointer -- in layman's terms, the PayID server gives you an address for the  payment-network you ask it for.

In that way, PayID can yield a Payment Network Address that can be brought to the OpenPayments server, and OP works roughly the same way for each -- though "[getting payment details](https://docs.openpayments.dev/payments)" becomes an optional step.

For example:

* When you bring an ILP-network address to OP, you need to do SPSP to get connection details. 
* In Crypto, you might want to get info about your non-custodial wallet, or do nothing. 
* In ACH you do nothing.

documentation/docs/payid.mdx
```diff
@@ -0,0 +1,74 @@
...
+   2. The URL's host is **host**
+   3. The URL's path is **path**
+
+For example, the PayID `alice$wallet.example` resolves to the URL
+`https://wallet.example/$/alice`.
+
+The `/$/` path prefix minimizes the chance that Web servers will have path
+conflicts with their existing services.
+
+## Privacy
```
sappenin, line 63: Relating to PayID over OpenPayments, one thing my team has been exploring a lot is where it makes sense to get certain data in non-ILP open-payments flows (e.g.,  PayID + XRPL). E.g., sometimes it makes sense to put data into the OP layer, and sometimes it makes sense to defer to the PayID layer.

Probably something we should add to the agenda for Wednesday's call -- but in short I think in OP primitives, we're missing the concept of a "Payment Network" when it comes to something like an `invoice` or a `mandate`. 

For example, we have an `address` (PaymentPointer or XRPL addr or BTC addr or whatever); we have the `assetCode` and `assetScale`, but we're missing the underlying network.

As currently written, the OP spec assume ILP only, so the payment network is inferred throughout the protocol.

documentation/docs/payid.mdx
```diff
@@ -0,0 +1,74 @@
...
+
+A **PayID** is an alternative identifier which, unlike Payment Pointers that
+identify accounts, identifies [account owners](./terminology#account-owner).
+
+The
+[PayID URI specification](https://github.com/xpring-eng/rfcs/blob/master/payid/dist/spec/payid-uri.txt)
+defines the format of a PayID to be `acctpart "$" host`.
+
+With respect to Open Payments the `acctpart` is the account owner identifier and
+`host` is the origin of the Open Payments server.
```
nkramer44, line 31: We've seen a lot of partners host their PayID server at specific subdomains (ie payid.wallet.com), which results in PayIDs that look like `nkramer$payid.wallet.com`.  I think we want to make it clear that the `host` portion of a PayID will most likely **not** be the same as the host of the Open Payments server.

A likely scenario is that PayID and Open Payments operators will put a proxy in front of the two respective servers, and proxy requests to the correct server based on something in the request.  To reduce as much complexity in the proxy logic as possible, I think it makes sense to put Open Payments discovery at something like `/{acctPart}/.well-known/open-payments`.  This makes it abundantly clear to the proxy that that request is meant for the Open Payments server.  Once a client has done discovery on a PayID, they can do Open Payments things at specific account URLs.

For example, given a PayID of `nkramer$wallet.com`, a client can make a request to `https://wallet.com/nkramer/.well-known/open-payments`, which will be proxied to `https://openpayments.wallet.com/nkramer/.well-known/open-payments`.  This will return account URLs such as `https://openpayments.wallet.com/nkramerXRPaccount`.

documentation/docs/payments.mdx
adrianhopebailie: > PayID server treat sometihng like an XRPL address on-par with a Payment Pointer -- in layman's terms, the PayID server gives you an address for the payment-network you ask it for.

That's not really how Open Payments works though. A Payment Pointer is not enough info to use to send money to someone. It's an identifier that you use to discover that address (the ILP Address).

None of the primitives in Open Payments are payment network specific, the only assumption we have is that if you start with a Payment Pointer then you're guaranteed to be able to get an ILP address for that account.

Payment Pointers and PayIDs are interchangeable, they only differ in that a PP identifies an account that is guaranteed to be accessible via ILP and a PayID identifies an account owner that may have multiple accounts on various networks. Both are used as identifiers to follow a discovery flow and ultimately resolve an account and payment network address(es) for that account.

documentation/docs/payid.mdx
adrianhopebailie: I think this is a mis-use of the well-known address space which to my understanding is for hosting services per host. 

A PayID has two parts, the `host` and `account` so transposing it to a URL in the well-known space would mean something like:

`https://{host}/.well-known/payid/{account}`

Also, where I think a PayID is unique is that it's a human-friendly identifier and there is value in the URL form being human-friendly too.

I'm hopeful that wallets will actually host `text/html` representations of their owner resources that are something like an open protocol version of [Cash Tags](https://cash.app/help/us/en-gb/3123-cashtags)

As a wallet I'd much prefer my users to have a public page at `https://wallet.example/$/david` that is a direct transposition of their PayID than `https://wallet.example/.well-known/payid/david`

documentation/docs/payid.mdx
adrianhopebailie: > I think in OP primitives, we're missing the concept of a "Payment Network" when it comes to something like an invoice or a mandate.

I don't agree. The payment network is implied by the supported media types. As a client you can request an invoice or account resource and specify all the ways you are able to pay into that resource (i.e. all the networks you are on) and the response provides the payment details for that network embedded in the resource.

> For example, we have an address (PaymentPointer or XRPL addr or BTC addr or whatever); we have the assetCode and assetScale, but we're missing the underlying network.

A Payment Pointer is not an address, it's an identifier, like a PayID. The address is an ILP address.

> As currently written, the OP spec assume ILP only, so the payment network is inferred throughout the protocol.

Not true. OP mandates ILP as a supported network if the discovery was done from a Payment Pointer. This ensures backward compatibility with the existing ecosystem where Payment Pointers are always identifying ILP-enabled wallets.

If you have an account or invoice resource URL and you query it with other accepted media types you can get a payment address for other networks.

documentation/docs/payid.mdx
adrianhopebailie: > I think we want to make it clear that the host portion of a PayID will most likely not be the same as the host of the Open Payments server.

Why? What are the functions of a PayID server that they need to be hosted independently of an Open Payments server?

nkramer44: Imagine a wallet is running Xpring's PayID server implemented in JS, and running our OPS which is implemented in Java.  I'm not sure how you would host two different runtimes on the same host without placing a proxy in front of both of them. 

My point is that in order for your PayID to allow OP discovery, there would need to be a proxy sitting in front of both the PayID server and the OP server.  Without a defined OP Discovery namespace, the proxy's routing logic becomes really complicated.

sappenin: >  the proxy's routing logic becomes really complicated.

I will second this feedback. I think we're _wildly_ underestimating how difficult it will be for any mature organization to "just deploy a proxy" for this kind of stuff.

Two trivial examples: The last company I worked for suffered outages due to proxy-config complexity and had very strict rules about **not** putting API business logic into the edges. They tried _very hard_ to keep this logic in the API applications themselves because this was the best place for that logic to live (especially from a QA perspective). Getting something like this into their proxy/LB config would have taken months, and lots of $$ for all the teams to collaborate around the changes because proxy changes tend to be organizationally-cross-cutting (i.e., everyone has a thing in the proxy). Not to mention that the proxy changes were always high-risk because there wasn't a single QA system in-place for this layer of the infrastructure, because it was cross-cutting. In short, the rule was "don't put stuff into the edge layer"

Another example that comes to mind is "any large company". E.g., imagine if you had to deploy this to https://ibm.com. How long do you think their "just tweak the proxy" process is?

adrianhopebailie: I still don't understand why a wallet would run a PayID server and an Open Payments server? As I said above:

> What are the functions of a PayID server that they need to be hosted independently of an Open Payments server?

If I look at this through the lens of a wallet that already implements Open Payments then I see PayID as a new feature I am adding to my wallet.

So I would issue my users with a PayID and tell them they can use this to enable other wallets to pay them.

So then my users use their PayID somewhere and the other system (the Open Payments client) follows the discovery flow I've defined here to discover all the user's accounts and they use Open Payments to complete a transaction with them.

adrianhopebailie:
Gonna close this for now and wait and see how the PayID protocol evolves post-launch.

If we can get PayID to adjust slightly and support the concept of accounts and addresses as different entities then we may not need a stand-alone discovery protocol.

PR #51, opened 2020-11-11
Fix JSON syntax in accounts.mdx

michielbdejong:


michielbdejong:
I hand-edited this, haven't checked the resulting JSON, but it looks more correct to me this way.

kincaidoneil (approved):

PR #75, opened 2021-11-02
fix(invoices): invoice expiresAt is required

sentientwaffle:


adrianhopebailie:
Closing as this relates to older documentation that is no longer being maintained

PR #88, opened 2022-02-16
Open Payments v2

wilsonianb:
I'm opening a draft pull request for easier inline discussion.


notes/README.md
```diff
@@ -0,0 +1,592 @@
...
+5. Expired
+6. Failed
+
+By default a new payment will have the state of **Pending**. A client with permission to authorize the payment can update the state to **Authorized** following which the ASPSP will attempt to execute the payment and the client can no longer update the payment.
+
+When creating a new payment the client must provide at least one of, a value for the **sendAmount**, a value for the **receiveAmount**, or an incoming payment URL as the value for the **receiver**.
+
+If the client provides an incoming payment URL the ASPSP must query it to determine the **receiveAmount**. If the ASPSP is unable to get a **receiveAmount** from the URL provided (insufficient permissions, or the URL is not for a valid incoming payment in an appropriate state) then the ASPSP should set the state of the Outgoing Payment to **Failed**.
+
+If the client provides a **receiveAmount** or incoming iayment then the ASPSP SHOULD calculate a **sendAmount** that will be debited from the underlying account and set this in the response.
```
wilsonianb, line 302:
```suggestion
If the client provides a **receiveAmount** or incoming payment then the ASPSP SHOULD calculate a **sendAmount** that will be debited from the underlying account and set this in the response.
```

wilsonianb: [Rafiki](https://github.com/interledger/rafiki/blob/main/docs/transaction-api.md#outgoing-payment-lifecycle) currently calculates the outgoing payment quote asynchronously after the payment is created because the quoting process isn't guaranteed to complete before the create outgoing payment request times out.
Maybe this could also have an optional `Quoting` state before the payment is `Pending`?

notes/README.md
```diff
@@ -0,0 +1,592 @@
...
+| receivingAccount | Maybe | Either a receivingAccount or a receivingPayment MUSt be provided.
+| receivingPayment | Maybe | Either a receivingAccount or a receivingPayment MUSt be provided.
+| receiveAmount | Maybe | Required if no sendAmount and no receivingPayment is provided. assetCode and assetScale MUST match those of the receivingAccount.
+| sendAmount | Maybe | Required if no receiveAmount and no receivingPayment is provided.  assetCode and assetScale MUST match those of the underlying Account.
+| description | No | A description or message that will be available to the sender and MAY be used by the provider if creating an incoming payment at the receivingAccount.
+| externalRef | No | An external reference number or code that can be used to reconcile the payment with external systems such as an invoice number.
+| receiptNonce | No | A random nonce to identify the underlying ILP STREAM connection used to send this payment. (Optional)
+| receiptSecret | No | A secret used to sign receipts returned on the ILP STREAM connection used to send this payment. (Optional)
```
wilsonianb, line 363:
```suggestion
```
Generally the sender shouldn't have access to the receipt secret, otherwise they could produce bogus receipts.
Is there a use case for this being specified in the outgoing payment?

notes/README.md
```diff
@@ -0,0 +1,592 @@
...
+
+If the receiver is an **Account** resource then a new **sendAmount** OR **receiveAmount** can be provided.
+
+If the current state of the payment is **Expired** then the state must also be updated to **Pending**.
+
+The request body contains an Outgoing Payment resource encoded as JSON with only the following properties: 
+
+| Property | Required | Notes
+| -- | -- | --
+| state | Maybe | If the current state is pending or then this MUST have a value of pending or not be provided. If the current state is expired or then this MUST have a value of pending. If the current state is anything but expired or pending then this request MUST be rejected.
```
wilsonianb, line 392:
```suggestion
| state | Maybe | If the current state is pending then this MUST have a value of pending or not be provided. If the current state is expired or then this MUST have a value of pending. If the current state is anything but expired or pending then this request MUST be rejected.
```
?

notes/README.md
```diff
@@ -0,0 +1,592 @@
...
+### Actions
+
+Actions specify the actions a client can perform against a resource. Different actions are available for different grant types.
+
+*__NOTE:__ We need to validate the account related actions with use cases before we can define the access control properly.*
+
+Incoming Payment
+* Create
+* Read
+* RequestReceipts
```
wilsonianb, line 453:
```suggestion
```

adrianhopebailie: @wilsonianb - I'm struggling to keep this all up to date here. Can we discuss on HackMD until things stabilise a bit?

https://hackmd.io/@openpayments2

wilsonianb: :+1: I'll close this for now in favor of that.

PR #96, opened 2022-04-05
chore: updating as discussed

sabineschaller:
- rename `amount` -> `value`
- make `value` of type string with regex
- make `assetCode` and `assetScale` required
- add `createdAt` and `updatedAt` to incoming and outgoing payment GET response
- fix outgoing payment PUT

### DIFF
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -1013,11 +1013,11 @@ components:
               authorized: true
               receivingAccount: 'https://openpayments.guide/aplusvideo/'
               receiveAmount:
-                amount: 2500
+                value: 2500
                 assetCode: USD
                 assetScale: 2
               sendAmount:
-                amount: 2500
+                value: 2500
                 assetCode: USD
                 assetScale: 2
               fundsAvailable: true
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -707,7 +707,7 @@ components:
         incomingAmount:
           $ref: '#/components/schemas/amount'
         receivedAmount:
-          $ref: '#/components/schemas/receivedAmount'
+          $ref: '#/components/schemas/amount'
         expiresAt:
           type: string
           description: The date and time when payments into the incoming payment will no longer be accepted.
@@ -856,7 +856,7 @@ components:
       description: |-
         All amounts in open payments are represented as a value and an asset code and scale.
 
-        The `value` is a positive monetary value, represented as a string. The following regular expression is an implementation of this definition: `^[1-9]\d*$`.
+        The `value` is an unsigned 64-bit integer amount, represented as a string.
 
         The `assetCode` is a code that indicates the underlying asset. In most cases this SHOULD be a 3-character ISO 4217 currency code.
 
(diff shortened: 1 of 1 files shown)


open-api-spec.yaml
```diff
@@ -810,32 +840,43 @@ components:
...
+          pattern: '2022-04-01T10:24:36.11Z'
+          description: The date and time when the outgoing payment was updated.
     amount:
       title: Amount
       type: object
       description: |-
-        All amounts in open payments are represented as an unsigned 64-bit integer amount and an asset code and scale.
+        All amounts in open payments are represented as a value and an asset code and scale.
+
+        The `value` is a positive monetary value, represented as a string. The following regular expression is an implementation of this definition: `^[1-9]\d*$`.
```
wilsonianb, line 859:
```suggestion
        The `value` is an unsigned 64-bit integer amount, represented as a string. The following regular expression is an implementation of this definition: `^[1-9]\d*$`.
```

open-api-spec.yaml
```diff
@@ -810,32 +840,43 @@ components:
...
-        amount:
-          type: integer
+        value:
+          type: string
           description: The amount that must be paid into the receiver’s account.
-          example: 2500
-          maximum: 18446744073709552000
-          minimum: 0
+          example: '2500'
+          pattern: '^[1-9]\d*$'
```
wilsonianb, line 879: maybe limit it two 20 characters to help limit uint64 overflow?

open-api-spec.yaml
```diff
@@ -810,32 +840,43 @@ components:
...
           assetScale: 2
-        - amount: 123456789
+        - value: '123456789'
           assetCode: BTC
           assetScale: 8
       properties:
-        amount:
-          type: integer
+        value:
+          type: string
```
wilsonianb, line 876: Could we add a custom(?) `uint64` format like here:
https://developers.google.com/discovery/v1/type-format

open-api-spec.yaml
```diff
@@ -691,7 +707,7 @@ components:
         incomingAmount:
           $ref: '#/components/schemas/amount'
         receivedAmount:
-          $ref: '#/components/schemas/amount'
+          $ref: '#/components/schemas/receivedAmount'
```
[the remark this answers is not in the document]
sabineschaller: Because the received amount can have a value of 0 whereas all other amounts should have a value greater than 0. If we don't want to enforce that on the spec level, I can change that such that only one amount type is necessary.

open-api-spec.yaml
sabineschaller: Actually, we can.

open-api-spec.yaml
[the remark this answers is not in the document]
sabineschaller: I changed it to just 1 amount type with type string, format uint64.

wilsonianb (approved):

wilsonianb (commented):
The amounts in the `list-outgoing-payments` examples need to be updated:
https://github.com/interledger/open-payments/blob/7d17ed21611e362bae3416f59ba9119bfc95a382/open-api-spec.yaml#L1016-L1020

PR #97, opened 2022-04-06
feat: add quotes

wilsonianb:
Remove Outgoing Payment authorization and update.

### DIFF
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -12,10 +12,11 @@ info:
 
     All resource and collection resource representations use JSON and the media-type `application/json`.  
 
-    The `account` resource has two collections of sub-resources and a single **account holder** sub-resource:   
+    The `account` resource has three collections of sub-resources and a single **account holder** sub-resource:   
     1. `/incoming-payments` contains the **incoming payment** sub-resources  
     2. `/outgoing-payments` contains the **outgoing payment** sub-resources  
-    3. `/holder` contains the **outgoing payment** sub-resources  
+    3. `/quotes` contains the **quote** sub-resources  
+    4. `/holder` contains the **outgoing payment** sub-resources  
 
     Access to resources and permission to execute the methods exposed by the API is determined by the grants given to the client represented by an access token used in API requests.
   summary: An API for open access to financial accounts to send and receive payments.
@@ -253,14 +254,11 @@ paths:
                     id: 'https://openpayments.guide/alice/outgoing-payments/8c68d3cc-0a0f-4216-98b4-4fa44a6c88cf'
                     accountId: 'https://openpayments.guide/alice/'
                     state: prepared
-                    authorized: false
                     receivingAccount: 'https://openpayments.guide/bob/'
                     sendAmount:
                       amount: 2500
                       assetCode: USD
                       assetScale: 2
-                    fundsAvailable: true
-                    expiresAt: '2022-04-12T23:20:50.52Z'
                     description: Thank you for the shoes.
         '401':
           $ref: '#/components/responses/401'
@@ -270,6 +268,10 @@ paths:
         content:
           application/json:
             examples:
+              Create an outgoing payment based on a quote:
+                value:
+                  quoteId: 'https://openpayments.guide/alice/quotes/ab03296b-0c8b-4776-b94e-7ee27d868d4d'
+                  externalRef: INV2022-02-0137
               Create fixed-send-amount outgoing payment for $25 to pay invoice INV2022-02-0137:
                 value:
                   receivingAccount: 'https://openpayments.guide/alice'
@@ -285,10 +287,12 @@ paths:
             schema:
               type: object
               properties:
-                authorized:
-                  type: boolean
-                  description: Indicates if the payment is authorized to begin processing.
-                  default: false
+                quoteId:
+                  type: string
+                  format: uri
+                  description: The URL of the quote defining this payment's amounts.
+                  example: 'https://openpayments.guide/alice/quotes/ab03296b-0c8b-4776-b94e-7ee27d868d4d'
+                  readOnly: true
                 receivingAccount:
                   type: string
                   format: uri
@@ -301,14 +305,6 @@ paths:
                   $ref: '#/components/schemas/amount'
                 sendAmount:
                   $ref: '#/components/schemas/amount'
-                fundsAvailable:
-                  type: boolean
-                  description: Indicates that there are funds available in the underlying account to complete this payment. This is not a guarantee they will still be available when the payment is attempted but it is a good indicator to a client that it should attempt to authorize the payment.
-                  readOnly: true
-                expiresAt:
-                  type: string
-                  description: The date and time when the calculated `sendAmount` is no longer valid.
-                  readOnly: true
                 description:
                   type: string
                   description: Human readable description of the outgoing payment that will be visible to the account holder and shared with the receiver.
@@ -345,7 +341,6 @@ paths:
                         - id: 'https://openpayments.guide/alice/outgoing-payments/8c68d3cc-0a0f-4216-98b4-4fa44a6c88cf'
                           accountId: 'https://openpayments.guide/alice/'
                           state: completed
-                          authorized: true
                           receivingAccount: 'https://openpayments.guide/aplusvideo/'
                           receiveAmount:
                             amount: 2500
@@ -355,21 +350,16 @@ paths:
                             amount: 2500
                             assetCode: USD
                             assetScale: 2
-                          fundsAvailable: true
-                          expiresAt: '2022-04-12T23:20:50.52Z'
                           description: APlusVideo subscription
                           externalRef: 'customer: 847458475'
                         - id: 'https://openpayments.guide/alice/outgoing-payments/0cffa5a4-58fd-4cc8-8e01-7145c72bf07c'
                           accountId: 'https://openpayments.guide/alice/'
                           state: completed
-                          authorized: true
                           receivingPayment: 'https://openpayments.guide/shoeshop/2fe92c6f-ef0d-487c-8759-3784eae6bce9'
                           sendAmount:
                             amount: 7026
                             assetCode: USD
                             assetScale: 2
-                          fundsAvailable: true
-                          expiresAt: '2022-04-30T14:20:32.49Z'
                           description: Thank you for your purchase at ShoeShop!
                           externalRef: INV2022-8943756
                 backward pagination:
@@ -383,20 +373,16 @@ paths:
                         - id: 'https://openpayments.guide/alice/outgoing-payments/0cffa5a4-58fd-4cc8-8e01-7145c72bf07c'
                           accountId: 'https://openpayments.guide/alice/'
                           state: completed
-                          authorized: true
                           receivingPayment: 'https://openpayments.guide/shoeshop/2fe92c6f-ef0d-487c-8759-3784eae6bce9'
                           sendAmount:
                             amount: 7026
                             assetCode: USD
                             assetScale: 2
-                          fundsAvailable: true
-                          expiresAt: '2022-04-30T14:20:32.49Z'
                           description: Thank you for your purchase at ShoeShop!
                           externalRef: INV2022-8943756
                         - id: 'https://openpayments.guide/alice/outgoing-payments/8c68d3cc-0a0f-4216-98b4-4fa44a6c88cf'
                           accountId: 'https://openpayments.guide/alice/'
                           state: completed
-                          authorized: true
                           receivingAccount: 'https://openpayments.guide/aplusvideo/'
                           receiveAmount:
                             amount: 2500
@@ -406,8 +392,6 @@ paths:
                             amount: 2500
                             assetCode: USD
                             assetScale: 2
-                          fundsAvailable: true
-                          expiresAt: '2022-04-12T23:20:50.52Z'
                           description: APlusVideo subscription
                           externalRef: 'customer: 847458475'
         '401':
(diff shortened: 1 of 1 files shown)


open-api-spec.yaml
```diff
@@ -431,6 +415,142 @@ paths:
           in: query
           name: cursor
           description: The cursor key to list from
+  /quotes:
+    post:
+      summary: Create a Quote
+      tags: []
+      operationId: create-quote
+      responses:
+        '202':
```
wilsonianb, line 424: Should a quote have `state` instead of returning 202 before it's ready?

open-api-spec.yaml
```diff
@@ -431,6 +415,142 @@ paths:
...
+                  receivingPayment: 'https://openpayments.guide/alice/incoming-payments/37a0d0ee-26dc-4c66-89e0-01fbf93156f7'
+                  receiveAmount:
+                    amount: 2500
+            schema:
+              type: object
+              properties:
+                receivingAccount:
+                  type: string
+                  format: uri
+                  description: The URL of the account that is being paid.
```
wilsonianb, line 455: Should a quote auto-create the corresponding incoming payment?
I think ilp-pay would need to be updated to be able to quote for a `receivingAccount` without creating an incoming payment.

open-api-spec.yaml
```diff
@@ -285,10 +287,12 @@ paths:
             schema:
               type: object
               properties:
-                authorized:
-                  type: boolean
-                  description: Indicates if the payment is authorized to begin processing.
-                  default: false
+                quoteId:
```
wilsonianb, line 290: Should an outgoing payment for a quote be created via the quote url instead?
```
POST https://openpayments.guide/alice/quotes/ab03296b-0c8b-4776-b94e-7ee27d868d4d/outgoing-payment
```

open-api-spec.yaml
```diff
@@ -12,10 +12,11 @@ info:
 
     All resource and collection resource representations use JSON and the media-type `application/json`.  
 
-    The `account` resource has two collections of sub-resources and a single **account holder** sub-resource:   
+    The `account` resource has three collections of sub-resources and a single **account holder** sub-resource:   
     1. `/incoming-payments` contains the **incoming payment** sub-resources  
     2. `/outgoing-payments` contains the **outgoing payment** sub-resources  
-    3. `/holder` contains the **outgoing payment** sub-resources  
+    3. `/quotes` contains the **quote** sub-resources
```
wilsonianb, line 18: Should a quote be an account sub-resource or could the source asset be included in a request to a general `POST /quotes` endpoint?

wilsonianb (commented):
Can only a single outgoing payment be created from a particular quote?

open-api-spec.yaml
```diff
@@ -793,6 +900,68 @@ components:
...
+          example: 'https://openpayments.guide/aplusvideo/'
+        receivingPayment:
+          type: string
+          format: uri
+          description: The URL of the incoming payment that would be paid.
+          example: 'https://openpayments.guide/aplusvideo/incoming-payments/45d495ad-b763-4882-88d7-aa14d261686e'
+        receiveAmount:
+          $ref: '#/components/schemas/amount'
+        sendAmount:
+          $ref: '#/components/schemas/amount'
```
wilsonianb, line 964: The quote should have either:
- [`paymentType`](https://github.com/interledgerjs/interledgerjs/tree/master/packages/pay#quote) to indicate fixed send vs fixed receive
- `maxSendAmount` (with `receiveAmount`) and `maxReceiveAmount` (with `sendAmount`)

open-api-spec.yaml
```diff
@@ -793,6 +900,68 @@ components:
...
+          receivingAccount: 'https://openpayments.guide/aplusvideo/'
+          receiveAmount:
+            amount: 2500
+            assetCode: USD
+            assetScale: 2
+          sendAmount:
+            amount: 2500
+            assetCode: USD
+            assetScale: 2
+          fundsAvailable: true
```
adrianhopebailie, line 927: Let's drop `fundsAvailable` as discussed. If there are no funds available then the quote should fail.

open-api-spec.yaml
wilsonianb: Auto-create for outgoing payment, not quote.

open-api-spec.yaml
sabineschaller: I think that is messy. I think an outgoing payment should always be created via the open payments url and `quote` is another field in the body.

open-api-spec.yaml
wilsonianb: added `amountType`
https://github.com/interledger/open-payments/pull/97/commits/a8f520302e3798f61968c57dc085959865ebe3d6

open-api-spec.yaml
wilsonianb: No state since quote request will be synchronous
https://github.com/interledger/open-payments/pull/97/commits/3f50d4881a4177cfe9b14366453d9fd95ba365a4

open-api-spec.yaml
```diff
@@ -850,6 +1001,89 @@ components:
...
+          example: 'https://openpayments.guide/aplusvideo/incoming-payments/45d495ad-b763-4882-88d7-aa14d261686e'
+        receiveAmount:
+          $ref: '#/components/schemas/amount'
+        sendAmount:
+          $ref: '#/components/schemas/amount'
+        amountType:
+          type: string
+          description: The type of quote amount.
+          enum:
+            - fixed-send
```
wilsonianb, line 1070: Would a client actually need to do a fixed-send quote before authorization?
The grant can't have `receiveAmount`, so why not just create the fixed-send outgoing payment and specify the (minimum) `receiveAmount`?

wilsonianb: :thinking: But then the client wouldn't have a way to check if funds are available for a fixed-send outgoing payment...

open-api-spec.yaml
```diff
@@ -451,6 +435,181 @@ paths:
...
+          $ref: '#/components/responses/403'
+      requestBody:
+        content:
+          application/json:
+            examples:
+              Create fixed-send-amount quote for $25:
+                value:
+                  receivingAccount: 'https://openpayments.guide/alice'
+                  sendAmount:
+                    value: '2500'
```
sabineschaller, line 480: I thought we wanted to make `assetCode` and `assetScale` required.

open-api-spec.yaml
```diff
@@ -451,6 +435,181 @@ paths:
...
+              Create fixed-receive-amount quote for $25:
+                value:
+                  receivingPayment: 'https://openpayments.guide/alice/incoming-payments/37a0d0ee-26dc-4c66-89e0-01fbf93156f7'
+                  receiveAmount:
+                    value: '2500'
+            schema:
+              type: object
+              properties:
+                receivingAccount:
```
sabineschaller, line 489: Why does the receiving account matter for the quote?

wilsonianb: As with outgoing payments, the client may specify either `receivingAccount` or `receivingPayment`.
When `receivingAccount` is specified, an outgoing payment will create an incoming payment (`receivingPayment`), but a quote will not.

open-api-spec.yaml
```diff
@@ -784,31 +948,31 @@ components:
...
+          format: uri
+          description: The URL of the quote defining this payment's amounts.
+          example: 'https://openpayments.guide/alice/quotes/ab03296b-0c8b-4776-b94e-7ee27d868d4d'
+          readOnly: true
         state:
           type: string
           description: The state of the payment as described above.
           enum:
             - pending
+            - processing
```
wilsonianb, line 971: Could `processing` be removed? Without authorization, the distinction between `pending` and `processing` seems implementation specific.
Maybe `processing` could also be removed from incoming payment state, since you can check the `receivedAmount` of a `pending` incoming payment.

wilsonianb: Correct, although an outgoing payment can alse be `failed`.

>Partial sent/received funds should not change the state?

Currently outgoing payment state changes to `processing` *before* starting to send, not after the first successfully sent packet (like incoming payments).
If we add a `sentAmount` to outgoing payments, we could similarly look at that for a `pending` payment, like I'm suggesting we do with an `pending` incoming payment's `receivedAmount`.

open-api-spec.yaml
wilsonianb: The quote needs a [destination](https://github.com/interledgerjs/interledgerjs/tree/master/packages/pay#quoteoptions) (a [ResolvedPayment](https://github.com/interledgerjs/interledgerjs/tree/master/packages/pay#resolvedpayment)), either an account or an incoming payment.

sabineschaller: Of course :see_no_evil:

open-api-spec.yaml
sabineschaller: As per our discussion yesterday, are you including the changes to remove `state`?

wilsonianb: https://github.com/interledger/open-payments/pull/97/commits/9a91081daf5250e49075a4e226e2afdd2b7611f8

open-api-spec.yaml
```diff
@@ -823,15 +1013,8 @@ components:
...
           $ref: '#/components/schemas/amount'
-        fundsAvailable:
-          type: boolean
-          description: Indicates that there are funds available in the underlying account to complete this payment. This is not a guarantee they will still be available when the payment is attempted but it is a good indicator to a client that it should attempt to authorize the payment.
-          readOnly: true
-        expiresAt:
-          type: string
-          description: The date and time when the calculated `sendAmount` is no longer valid.
-          example: '2022-04-12T23:20:50.52Z'
-          readOnly: true
```
wilsonianb, line 834: If creating an outgoing payment:
- required a `quoteId` of a created quote and/or 
- synchronously performed the quote before resolving the create outgoing payment request

the outgoing payment schema could be a little cleaner:

wilsonianb, line 834: If creating an outgoing payment:
- required a `quoteId` of a created quote and/or 
- synchronously performed the quote before resolving the create outgoing payment request

the outgoing payment schema could be a little cleaner:
- `receivingPayment`, `sendAmount`, `receiveAmount` would all be required
- no `receivingAmount`

wilsonianb, line 834: If creating an outgoing payment:
- required a `quoteId` of a created quote and/or 
- synchronously performed the quote before resolving the create outgoing payment request

the outgoing payment schema could be a little cleaner:
- `receivingPayment`, `sendAmount`, `receiveAmount` would all be required
- no `receivingAmount`

wilsonianb: Otherwise, I'm considering implementing separate quoted and un-quoted outgoing payments database tables (maybe one [inheriting](https://www.postgresql.org/docs/current/ddl-inherit.html) from the other).

wilsonianb: https://github.com/interledger/open-payments/pull/97/commits/4bb20e6ebdb196b36b77d76dc6609eb7f9175c36 assumes that a request to create an outgoing payment without a pre-existing quote creates a quote synchronously.

open-api-spec.yaml
```diff
@@ -850,6 +1030,78 @@ components:
           format: date-time
           pattern: '2022-04-01T10:24:36.11Z'
           description: The date and time when the outgoing payment was updated.
+      required:
+        - id
+        - accountId
```
wilsonianb, line 1035: Should `quoteId` be required/exposed?

open-api-spec.yaml
wilsonianb: Or we would it be preferable to quickly return `202` + `Location` header for a request to create an outgoing payment without specifying a `quoteId`?

open-api-spec.yaml
```diff
@@ -759,61 +954,53 @@ components:
...
           example: 'https://openpayments.guide/alice/outgoing-payments/8c68d3cc-0a0f-4216-98b4-4fa44a6c88cf'
           readOnly: true
         accountId:
           type: string
           format: uri
-          description: The URL of the account this payment is being made into.
+          description: The URL of the account from which this payment is sent.
           example: 'https://openpayments.guide/alice/'
           readOnly: true
-        state:
```
wilsonianb, line 796: Reconsidering removing `state`: https://github.com/interledger/open-payments/issues/104#issuecomment-1102859610

open-api-spec.yaml
```diff
@@ -451,6 +467,177 @@ paths:
...
+                    receiveAmount:
+                      value: '2198'
+                      assetCode: EUR
+                      assetScale: 2
+                    createdAt: '2022-03-12T23:20:50.52Z'
+                    expiresAt: '2022-04-12T23:20:50.52Z'
+        '401':
+          $ref: '#/components/responses/401'
+        '403':
+          $ref: '#/components/responses/403'
```
wilsonianb, line 501: What should the response be if Rafiki's request to wallet doesn't succeed (such as if the sender's account doesn't have sufficient funds for the quoted amount)? 403 or whatever code the wallet returns? (What should the wallet return if the user doesn't have funds?)

wilsonianb: We discussed returning `5XX`

open-api-spec.yaml
wilsonianb: fwiw, Rafiki is using the quote id as the outgoing payment id (to enforce one outgoing payment per quote)
https://github.com/interledger/rafiki/pull/326/files#diff-60554f1175c8f9259e6dc5c51a32a1cf44b6d07301f92fec8f316c63ae87e2ab

open-api-spec.yaml
wilsonianb: https://github.com/interledger/open-payments/pull/97/commits/b8eec71fa60d56e24dd416821a3a279cf42a2d70 makes it so that a created `quoteId` is required when creating an outgoing payment.

open-api-spec.yaml
```diff
@@ -451,6 +439,177 @@ paths:
...
+                receivingPayment:
+                  type: string
+                  format: uri
+                  description: The URL of the incoming payment that is being paid.
+                receiveAmount:
+                  $ref: '#/components/schemas/amount'
+                sendAmount:
+                  $ref: '#/components/schemas/amount'
+        description: |-
+          A subset of the quotes schema is accepted as input to create a new quote.
```
sabineschaller, line 508: Should we say which subset that is?

open-api-spec.yaml
```diff
@@ -12,10 +12,11 @@ info:
 
     All resource and collection resource representations use JSON and the media-type `application/json`.  
 
-    The `account` resource has two collections of sub-resources and a single **account holder** sub-resource:   
+    The `account` resource has three collections of sub-resources and a single **account holder** sub-resource:
```
sabineschaller, line 15: Line 9 needs to be updated too to include the quote resource

open-api-spec.yaml
```diff
@@ -12,10 +12,11 @@ info:
 
     All resource and collection resource representations use JSON and the media-type `application/json`.  
 
-    The `account` resource has two collections of sub-resources and a single **account holder** sub-resource:   
+    The `account` resource has three collections of sub-resources and a single **account holder** sub-resource:   
     1. `/incoming-payments` contains the **incoming payment** sub-resources  
     2. `/outgoing-payments` contains the **outgoing payment** sub-resources  
-    3. `/holder` contains the **outgoing payment** sub-resources  
+    3. `/quotes` contains the **quote** sub-resources  
+    4. `/holder` contains the **outgoing payment** sub-resources
```
sabineschaller, line 19: Do you want to remove the holder resource in your PR as well or should I create a new one for that?

wilsonianb: I can remove it.

sabineschaller (approved):
LGTM

PR #106, opened 2022-04-21
chore(auth): update grant access information

### LINKED ISSUE #103: Authorization Sever Spec updates
* limits in access objects mirror incoming and outgoing payment data model, respectively
  * remove `interval` property
* `startsAt` follows [ISO 8601 Repeating intervals](https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals)
* point out that all amounts are maxima, i.e. it is possible to  create many payments until the total equals the granted amount

sabineschaller:
closes #103 
Additinally, adds `id` parameter for token endpoints.

### DIFF
--- a/auth-server-open-api-spec.yaml
+++ b/auth-server-open-api-spec.yaml
@@ -290,8 +290,13 @@ paths:
         '404':
           description: Not Found
       description: Cancel a grant request.
-  /token:
-    parameters: []
+  '/token/{id}':
+    parameters:
+      - schema:
+          type: string
+        name: id
+        in: path
+        required: true
     post:
       summary: Rotate Access Token
       operationId: post-token
--- a/auth-server-open-api-spec.yaml
+++ b/auth-server-open-api-spec.yaml
@@ -72,10 +72,8 @@ paths:
                           - read
                           - update
                         identifier: 'https://openpayments.guide/alice'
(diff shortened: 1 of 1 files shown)


auth-server-open-api-spec.yaml
```diff
@@ -429,23 +429,11 @@ components:
...
-          description: ISO 8601 date-time indicating when the grant becomes effective.
-        expiresAt:
-          type: string
-          format: date-time
-          description: ISO 8601 date-time indicating when the grant expires.
-        interval:
-          type: string
-          format: duration
-          description: 'ISO 8601 duration indicating how often the amounts, `sendAmount` or `receiveAmount`, are available.'
         receivingAccount:
```
wilsonianb, line 437: The descriptions for `receivingAccount` and `receivingPayment` say that one is required. Is that still the case?

wilsonianb (commented):
Examples should also updated to remove `startsAt`, `expiresAt`, `interval`

auth-server-open-api-spec.yaml
sabineschaller: If there is neither, it means I can just create outgoing payments to whoever using that grant. Sounds like we need that for web monetization. So they should both be optional.

wilsonianb (approved):

PR #112, opened 2022-04-28
fix: include accountId in open-api-spec.yaml paths

wilsonianb:
https://spec.openapis.org/oas/v3.1.0#paths-object


wilsonianb:
oops, I didn't realize `servers.url` included the account id
https://github.com/interledger/open-payments/blob/0c923e8110281197b2e6cdbecf31e2fc87658ee0/open-api-spec.yaml#L22-L26

Will that need to change to support `https://wallet.com/INCOMING_PAYMENT_ID`?

PR #113, opened 2022-04-29
chore: remove deprecated example field

wilsonianb:
https://spec.openapis.org/oas/v3.1.0#schema-object
>The `example` property has been deprecated in favor of the JSON Schema `examples` keyword. Use of `example` is discouraged, and later versions of this specification may remove it.


sabineschaller (approved):
Thanks for fixing this!

PR #122, opened 2022-05-11
chore: define account id as path parameter

wilsonianb:
This is needed for request validation both now and when the account id is removed from the `/{accountId}/incoming-payments/{id}` path (in order for the spec to distinguish between paths that do and do not include the account id).

### DIFF
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -22,10 +22,7 @@ info:
   contact:
     email: tech@interledger.org
 servers:
-  - url: 'https://openpayments.guide/alice'
-    description: alice
-  - url: 'https://openpayments.guide/bob'
-    description: bob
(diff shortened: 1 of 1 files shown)


open-api-spec.yaml
```diff
@@ -36,7 +33,7 @@ tags:
   - name: quote
     description: quote operations
 paths:
-  /:
+  '/{id}':
```
sabineschaller, line 36: Why not also call this `accountId` like it is in the other endpoints?

wilsonianb: I was thinking of consistently making `id` be the resource of the particular endpoint, but I don't feel strongly about it.

sabineschaller: Tbh, me neither. We can have `id` be the resource of the endpoint.

Issue #104, opened 2022-04-13
Open Payment Spec update: * remove `state` in incoming and outgoing payments


sabineschaller:
* incoming payment: 
  * add `completed` flag
  * payment is expired if `completed=false` and `now>expiresAt`
  * if payment is not complete and not expired, it is `pending`
* outgoing payment: 
  * add `failed` flag
  * payment is complete if `sendAmount=sentAmount`
  * if payment has not failed and `sendAmount!=sentAmount`, payment is `pending`
 
@wilsonianb lmk if I summarized that correctly

wilsonianb:
>payment is complete if sendAmount=sentAmount

With quote amounts being fixed, outgoing payments are fixed-receive. So when sending completes, `sentAmount` is `<=sendAmount`. If the leftover payment balance from `sendAmount` is handled asynchronously (such as Rafiki's current [`outgoing_payment.completed`](https://github.com/interledger/rafiki/blob/main/docs/transaction-api.md#payment-resolution) webhook event), the outgoing payment wouldn't be technically be complete yet (`sentAmount < sendAmount`).
I'm wondering if either we need a `completed` flag (in addition to `failed`...) or we still have `state` (`pending`, `completed`, `failed`). Rafiki's implementation will need one of those two options internally.

wilsonianb:
:thinking: OTOH, would we every want to return an outgoing payment as `completed` (via either flag or `state`) with `sentAmount`<`sendAmount` (due to unresolved leftover balance)? Would an implementation always just return `sentAmount` as `sendAmount`? In which case `payment is complete if sendAmount=sentAmount` would still suffice.

wilsonianb:
>* incoming payment: 
>   * add `completed` flag

I keep thinking back to the idea of completing an incoming payment by making `incomingAmount` equal `receivedAmount`.
It would make it consistent with outgoing payments:
>payment is complete if sendAmount=sentAmount

I think there were concerns about knowing the precise `receivedAmount` value to use in the `PUT`. What' if it were instead `POST /{id}/complete`?

sabineschaller:
> I keep thinking back to the idea of completing an incoming payment by making `incomingAmount` equal `receivedAmount`.

But `incomingAmount` is optional.

> I think there were concerns about knowing the precise `receivedAmount` value to use in the PUT request body. What' if it were instead `POST /{id}/complete`?

The PUT only takes `completed: true` as body parameter. Which `receivedAmount` are you referring to?
[...]

PR #126, opened 2022-05-24
chore: add /complete endpoint and completed flag

wilsonianb:
- fixes https://github.com/interledger/open-payments/issues/104

### DIFF
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -723,6 +723,25 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/incoming-payment'
+              examples:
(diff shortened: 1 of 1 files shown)


open-api-spec.yaml
```diff
@@ -710,9 +710,12 @@ paths:
         in: path
         required: true
         description: Incoming Payment identifier
-    put:
-      summary: Update an Incoming Payment
-      operationId: update-incoming-payment
+  '/{accountId}/incoming-payments/{id}/complete':
```
sabineschaller, line 713: Should we add an example response body?

adrianhopebailie (approved):

PR #127, opened 2022-05-24
chore: add pay endpoint and `payId` property

sabineschaller:
- removes STREAM credentials from incoming payments
- adds `payId` property to incoming payments
- adds `/pay/{id}` endpoint to request STREAM credentials


sabineschaller:
We decided to go with another schema.

PR #129, opened 2022-05-31
fix: change list responses from array to object

wilsonianb:
- update [`pagination`](https://relay.dev/graphql/connections.htm#sec-undefined.PageInfo.Fields) list response field:
  - makes properties required
  - remove `first`/`last`
- update [pagination](https://relay.dev/graphql/connections.htm#sec-Arguments) query parameters
  - enforce mutually exclusive forward/backward pagination parameters
  - ~replace `cursor` with explicit `before`/`after`~

### DIFF
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -340,106 +340,104 @@ paths:
           content:
             application/json:
               schema:
-                type: array
-                items:
-                  $ref: '#/components/schemas/list-outgoing-payments'
+                $ref: '#/components/schemas/list-outgoing-payments'
               examples:
                 forward pagination:
                   value:
-                    - pagination:
-                        startCursor: 241de237-f989-42be-926d-c0c1fca57708
-                        endCursor: 315581f8-9967-45a0-9cd3-87b60b6d6414
-                        hasNextPage: true
-                        first: 10
-                      result:
-                        - id: 'https://openpayments.guide/alice/outgoing-payments/8c68d3cc-0a0f-4216-98b4-4fa44a6c88cf'
-                          accountId: 'https://openpayments.guide/alice/'
-                          failed: false
-                          receivingPayment: 'https://openpayments.guide/aplusvideo/incoming-payments/45d495ad-b763-4882-88d7-aa14d261686e'
-                          receiveAmount:
-                            value: '2500'
-                            assetCode: USD
-                            assetScale: 2
-                          sendAmount:
-                            value: '2500'
-                            assetCode: USD
-                            assetScale: 2
-                          sentAmount:
-                            value: '2500'
-                            assetCode: USD
-                            assetScale: 2
-                          createdAt: '2022-03-12T23:20:50.52Z'
-                          updatedAt: '2022-04-01T10:24:36.11Z'
-                          description: APlusVideo subscription
-                          externalRef: 'customer: 847458475'
-                        - id: 'https://openpayments.guide/alice/outgoing-payments/0cffa5a4-58fd-4cc8-8e01-7145c72bf07c'
-                          accountId: 'https://openpayments.guide/alice/'
-                          failed: false
-                          receivingPayment: 'https://openpayments.guide/shoeshop/2fe92c6f-ef0d-487c-8759-3784eae6bce9'
-                          sendAmount:
-                            value: '7026'
-                            assetCode: USD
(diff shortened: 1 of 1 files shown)


open-api-spec.yaml
```diff
@@ -340,106 +340,111 @@ paths:
...
-                            assetScale: 2
-                          createdAt: '2022-03-12T23:20:50.52Z'
-                          updatedAt: '2022-04-01T10:24:36.11Z'
-                          description: Thank you for your purchase at ShoeShop!
-                          externalRef: INV2022-8943756
+                    pagination:
+                      startCursor: 241de237-f989-42be-926d-c0c1fca57708
+                      endCursor: 315581f8-9967-45a0-9cd3-87b60b6d6414
+                      hasNextPage: true
+                      first: 10
```
sabineschaller, line 358: Didn't you also want to remove the `first` and `last`? They are not really necessary since once can just get the length of the result.

sabineschaller (commented):
Good catch!

open-api-spec.yaml
```diff
@@ -340,127 +324,122 @@ paths:
           content:
             application/json:
               schema:
-                type: array
-                items:
-                  $ref: '#/components/schemas/list-outgoing-payments'
+                type: object
+                properties:
+                  pagination:
```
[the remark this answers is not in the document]
wilsonianb: I've dropped the pagination query parameter changes from this pr to contain it pagination response.
I'll open a second pr that implements :point_up_2:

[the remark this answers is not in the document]
sabineschaller: I feel like wherever you look, that is done differently.  

I've seen `count` instead of `first`/`last` but like I said before, that can be derived from the list of results anyways, so it's not really necessary. 

I do think, however, that  `hasPreviousPage`/`hasNextPage` add value so I would vote to keep those. 

I don't have strong feelings for `startCursor` and `endCursor` because they, also, can be derived from the results. It's just a bit more complicated.

I don't like `<resource-name>s` instead of `result` because responses have to be treated differently. But maybe that is not a problem :thinking:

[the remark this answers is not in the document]
wilsonianb: `count` would only exist as a query parameter.
So, the query params would change from `first`/`last` + `cursor` to `after`/`before` + `count`

I've seen things (including the linked post) recommending opaque cursors, so I'm leaning toward not having the client derive them from the results.
Maybe the `pageInfo`/`pagination` in the response could be:
```
  cursor: {
    after: <opaque-cursor>
    hasNextPage: true
  }
  ...
```
```
  cursor: {
    before: <opaque-cursor>
    hasPreviousPage: true
  }
  ...
```
I don't have a strong opinion on what we call the field that includes the list.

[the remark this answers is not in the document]
sabineschaller: > So, the query params would change from `first`/`last` + `cursor` to `after`/`before` + `count`

Yes, I've seen this combination. I'm happy either way. The changes shouldn't take too long. 

I like the `cursor` objects. I also don't have strong feelings about the name of the field with the list.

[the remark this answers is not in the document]
wilsonianb: I've opened an issue to handle it in a separate pr:
https://github.com/interledger/open-payments/issues/136

open-api-spec.yaml
```diff
@@ -1370,106 +1388,6 @@ components:
               updatedAt: '2022-04-01T10:24:36.11Z'
               description: 'Hi Mo, this is for the cappuccino I bought for you the other day.'
               externalRef: Coffee w/ Mo on 10 March 22
-    list-outgoing-payments:
```
sabineschaller, line 1373: Why is this removed but we still have `list-incoming-payments`?

wilsonianb: For some reason I was thinking `list-outgoing-payments` had a separate incoming payment schema without the stream credentials,
It _seems_ like `list-incoming-payments` could also be removed.

sabineschaller: The only reason to keep it is if we want to enforce that STREAM credentials are never part of the list response, not even the connection url.

sabineschaller: But then we shouldn't reference the incoming payment model

sabineschaller: PR: #137

PR #138, opened 2022-06-08
Big Cleanup

adrianhopebailie:
- Remove old Docusaurus website (code behind docs.openpayments.dev)
- Update all dependencies for landing page (now at openpayments.guide)
- Clean up landing page site to work with new versions on next.js and tailwind
- Add deployment action to push to GH pages


.github/workflows/openapi.yaml
```diff
@@ -17,6 +17,7 @@ jobs:
 
       - name: Validate open-api-spec.yaml
         run: |
+          echo "{\"extends\":[\"spectral:oas\",\"spectral:asyncapi\"]}" >> .spectral.json
```
sabineschaller, line 20: Why is that needed?

AlexLakatos (approved):
LGMT 🧹

.github/workflows/openapi.yaml
adrianhopebailie: It's required by the validator.
I removed the file that we had in the root of the repo in 3afb26ede2ac6c5c943ae712baf5198e914d3be9

PR #144, opened 2022-08-16
fix: add quote, fix outgoing, drop incoming limits

adrianhopebailie:


sabineschaller (approved):
LGTM

sabineschaller (approved):

sabineschaller (changes requested):
Please add the restriction of just one item in `access` also in the grant update request (`PATCH /continue/{id}`)

sabineschaller:
- remove `description` and `externalRef` from outgoing limits
- only allow interval in outgoing payment grant if the limits include an amount
- move `interval` into `limits`

auth-server-open-api-spec.yaml
```diff
@@ -516,6 +518,27 @@ components:
...
       description: The access associated with the access token is described using objects that each contain multiple dimensions of access.
+    limits-outgoing-recurring:
+      title: limits-outgoing-recurring
+      description: Open Payments specific property that defines the limits under which outgoing payments can be created.
+      type: object
+      properties:
+        receiver:
+          type: string
+          format: uri
+          description: URL for the incoming payment or ILP STREAM Connection.
```
wilsonianb, line 529: This could be made its own schema to avoid repeating.
Ditto for `interval`

auth-server-open-api-spec.yaml
```diff
@@ -516,6 +518,27 @@ components:
...
+        receiveAmount:
+          $ref: '#/components/schemas/amount'
+        interval:
+          type: string
+          description: '[ISO8601 repeating interval](https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals)'  
+      required: 
+        -  interval
+        -  oneOf
+          - sendAmount
+          - receiveAmount
```
wilsonianb, line 541: Does this allow both amounts?
https://stackoverflow.com/questions/31839578/how-to-define-a-json-schema-that-requires-at-least-one-of-many-properties/31841897#31841897

sabineschaller:
Please make grant updating impossible.

auth-server-open-api-spec.yaml
adrianhopebailie: Do we allow both amounts?

auth-server-open-api-spec.yaml
adrianhopebailie: Deferred to #156

adrianhopebailie:
> Please make grant updating impossible.

Deferred to https://github.com/interledger/open-payments/issues/157

PR #145, opened 2022-08-16
chore: rename accountId field to paymentPointer

wilsonianb:


open-api-spec.yaml
```diff
@@ -32,10 +32,10 @@ info:
   contact:
     email: tech@interledger.org
 servers:
-  - url: 'https://openpayments.guide'
+  - url: 'https://openpayments.guide/alice'
```
sabineschaller, line 35: I think we should keep `https://openpayments.guide` as URL and then add `payment-pointer` as a parameter.

wilsonianb: Unlike the account which was referred to by its id (as a path parameter), the payment pointer is the full base url (as demonstrated in the `servers` list)

wilsonianb: which is more consistent than before.
Previously `accountId` was both the uuid as a path parameter and the full account url as a field on all of the subresources.

wilsonianb: :thinking: Actually...
I'm not sure how to reconcile this with the [`/connections/{connectionId}` endpoint](https://github.com/interledger/open-payments/blob/main/open-api-spec.yaml#L85), which should be on the open payments server (not the payment pointer).

sabineschaller: Well.. yesterday we discussed potentially getting rid of `/connections`....

wilsonianb: I'm thinking we'd need to split [`open-api-spec.yaml`](https://github.com/interledger/open-payments/blob/main/open-api-spec.yaml) into two files:
- `payment-pointer-spec.yaml`? (everything except `/connections`)
- `connections-spec.yaml`?

wilsonianb: After discussing with @sabineschaller, I've reverted the servers list to not include payment pointer urls since they do not include the connections endpoints. We can consider splitting the files later if I run into issues implementing this change in Rafiki.

open-api-spec.yaml
```diff
@@ -45,20 +45,20 @@ tags:
   - name: stream-connection
     description: interledger STREAM connections
 paths:
-  '/{accountId}':
+  '/':
```
sabineschaller, line 48: Shouldn't that be `/{paymentPointer}`? And then all the other endpointes `/{paymentPointer}/...`, except connection, which is `/{connection}`?

wilsonianb: We'd still like the `servers` at the top to represent the payment pointer.
@matdehaast discovered that you can override the `servers` for a particular path, which we can do for `/connections/{connectionId}`

matdehaast (approved):

PR #147, opened 2022-08-19
feat: add quote to access schema

njlie:
The lack of the `quote` access type in the OpenAPI schema is blocking token introspection for Quote grants.

### DIFF
--- a/auth-server-open-api-spec.yaml
+++ b/auth-server-open-api-spec.yaml
@@ -849,7 +849,6 @@ components:
             enum:
               - read
               - create
-              - list
         identifier:
           type: string
           description: A string identifier indicating a specific resource at the RS.
--- a/auth-server-open-api-spec.yaml
+++ b/auth-server-open-api-spec.yaml
@@ -514,7 +514,7 @@ components:
         - $ref: '#/components/schemas/access-account'
         - $ref: '#/components/schemas/access-incoming'
         - $ref: '#/components/schemas/access-outgoing'
-        - $ref: '#components/schemas/access-quote'
+        - $ref: '#/components/schemas/access-quote'
       description: The access associated with the access token is described using objects that each contain multiple dimensions of access.
     limits-outgoing:
       title: limits-outgoing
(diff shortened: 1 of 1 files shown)


wilsonianb:
The OpenAPI errors should be resolved either here or in a separate pr.
I'm guessing some are pre-existing from when the [GitHub actions weren't running](https://github.com/interledger/open-payments/pull/146).

auth-server-open-api-spec.yaml
```diff
@@ -830,6 +832,33 @@ components:
...
+          enum:
+            - quote
+        actions:
+          type: array
+          description: The types of actions the client will take at the RS as an array of strings.
+          items:
+            type: string
+            enum:
+              - read
+              - create
```
wilsonianb, line 851: Do we want to allow listing quotes?

auth-server-open-api-spec.yaml
```diff
@@ -830,6 +832,33 @@ components:
...
+        identifier:
+          type: string
+          description: A string identifier indicating a specific resource at the RS.
+          format: uri
+        interval:
+          type: string
+          description: '[ISO8601 repeating interval](https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals)'
```
wilsonianb, line 858: `interval` is only meaningful with amount limits
```suggestion
```

auth-server-open-api-spec.yaml
sabineschaller: What would be the use case?

AlexLakatos: The OP spec has an [endpoint for listing quotes](https://docs.openpayments.guide/reference/list-quotes), I'm curious what the use case for that was initially.

sabineschaller: That may have been a copy-paste error

njlie: I didn't think there was one, but I deferred to the spec after seeing this. It seems like we should remove it(?)

sabineschaller: Let's remove it.

dclipp (approved):

PR #148, opened 2022-08-19
feat(OP-spec): remove min, max, default on pagination parameters

sabineschaller:
Min, max, and default values should be business logic.

### DIFF
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -1,7 +1,7 @@
 openapi: 3.1.0
 info:
   title: Open Payments
-  version: "1.0"
+  version: '1.0'
   license:
     name: Apache 2.0
     identifier: Apache-2.0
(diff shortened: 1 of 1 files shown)


open-api-spec.yaml
```diff
@@ -1290,19 +1298,15 @@ components:
       examples:
         - first: 10
           cursor: 315581f8-9967-45a0-9cd3-87b60b6d6414
+      additionalProperties: false
       properties:
         first:
           type: integer
-          minimum: 1
```
wilsonianb, line 1296: Maybe the `minimum` should stay

open-api-spec.yaml
```diff
@@ -1290,19 +1298,15 @@ components:
...
-          maximum: 100
-          default: 10
           description: The number of items to return.
         cursor:
           type: string
           minLength: 1
           description: The cursor key to list from.
-      required:
-        - first
```
wilsonianb, line 1305: I think `last` also shouldn't be required for `backward-pagination`

open-api-spec.yaml
sabineschaller: Sure, we can keep that.

wilsonianb (approved):

PR #161, opened 2022-08-30
chore: enforce connection inclusion in incoming payments

### LINKED ISSUE #160: Make incoming payment ilpStreamConnection type explicit
Enforce that `ilpStreamConnection` is:
- [`ilp-stream-connection`](https://github.com/interledger/open-payments/blob/main/open-api-spec.yaml#L918) in the create and read incoming payment responses
- `string` url in the list incoming payment responses
- not present in the `/complete` response

wilsonianb:
- fixes #160

### DIFF
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -135,7 +135,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/incoming-payment'
+                $ref: '#/components/schemas/incoming-payment-with-connection'
               examples:
                 New Incoming Payment for $25:
                   value:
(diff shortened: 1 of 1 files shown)


sabineschaller:
Should the connection be included in a list item if that incoming payment has been already completed?

wilsonianb:
https://github.com/interledger/open-payments/pull/161/commits/4a5060d5c673ec2f114c543a4ea27c870a77190d makes `ilpStreamConnection` optional, but now I'm wondering the incoming payment `completed` field could be removed in favor of `ilpStreamConnection` being present.

sabineschaller:
While it is possible, I don't think we should rely on `ilpStreamConnection` to infer the state of the payment. I think having it as an explicit field adds value.

sabineschaller (approved):

PR #174, opened 2022-09-09
chore: Add regex pattern for receiver

### LINKED ISSUE #156: Extract interval and receiver into schemas in auth spec

### LINKED ISSUE #172: Add regular expression pattern to receiver
Enforce the quote/outgoing payment (including grant limit) `receiver` field should be `https://` and include either `/incoming-payments/{uuid}` or `/connections/{uuid}`

wilsonianb:
Create receiver and interval schemas

- fixes #172 
- fixes #156

### DIFF
--- a/auth-server-open-api-spec.yaml
+++ b/auth-server-open-api-spec.yaml
@@ -565,16 +565,13 @@ components:
       type: object
       properties:
         receiver:
-          type: string
-          format: uri
-          description: URL for the incoming payment or ILP STREAM Connection.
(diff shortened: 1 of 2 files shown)


auth-server-open-api-spec.yaml
```diff
@@ -615,6 +609,19 @@ components:
         - value
         - assetCode
         - assetScale
+    interval:
```
sabineschaller, line 612: Should that include examples, too?

sabineschaller (approved):

PR #179, opened 2022-09-15
fix spec syntax

omertoast:
I came across a syntax error from this line while using [openapi-typescript](https://github.com/drwpow/openapi-typescript)

### DIFF
--- a/auth-server-open-api-spec.yaml
+++ b/auth-server-open-api-spec.yaml
@@ -577,7 +577,9 @@ components:
           description: '[ISO8601 repeating interval](https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals)'
       required:
         - interval
-        - oneOf - sendAmount - receiveAmount
+        - oneOf
+        - sendAmount
+        - receiveAmount
     limits-outgoing:
       title: limits-outgoing
       description: Open Payments specific property that defines the limits under which outgoing payments can be created.


wilsonianb:
Previous discussion: https://github.com/interledger/open-payments/pull/168#discussion_r961773935

auth-server-open-api-spec.yaml
```diff
@@ -577,7 +577,9 @@ components:
           description: '[ISO8601 repeating interval](https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals)'
       required:
         - interval
-        - oneOf - sendAmount - receiveAmount
+        - oneOf
+        - sendAmount
+        - receiveAmount
```
wilsonianb, line 582: What do we think of:
```
anyOf:
  - required:
    - interval
    - sendAmount
  - required:
    - interval
    - receiveAmount
```
https://stackoverflow.com/questions/41172240/json-schema-how-to-make-anyof-two-or-more-properties-required

wilsonianb: :point_up_2: that didn't work.

:point_down: this should:
- https://github.com/interledger/open-payments/pull/180

wilsonianb:
> I came across a syntax error from this line while using [openapi-typescript](https://github.com/drwpow/openapi-typescript)

How are you planning to use the generated types?
I had previously (unsuccessfully) tried to generate types using [oats](https://github.com/smartlyio/oats), but it looks like `openapi-typescript` works with our `open-api-spec.yaml`. :tada: 
I might look into replacing our manually defined types in Rafiki (example: https://github.com/interledger/rafiki/blob/e96d44e692660babd4e6c45d6e36178e9ee01f86/packages/backend/src/open_payments/payment/incoming/model.ts#L248-L260)

auth-server-open-api-spec.yaml
omertoast: I don't know the correct spec, I was just trying to get syntax work so yours is probably the right one.

omertoast:
> > I came across a syntax error from this line while using [openapi-typescript](https://github.com/drwpow/openapi-typescript)
> 
> How are you planning to use the generated types? I had previously (unsuccessfully) tried to generate types using [oats](https://github.com/smartlyio/oats), but it looks like `openapi-typescript` works with our `open-api-spec.yaml`. 🎉 I might look into replacing our manually defined types in Rafiki (example: https://github.com/interledger/rafiki/blob/e96d44e692660babd4e6c45d6e36178e9ee01f86/packages/backend/src/open_payments/payment/incoming/model.ts#L248-L260)

I'm creating a request validator service for the [rafiki-auth](https://github.com/omertoast/rafiki-auth) package @adrianhopebailie and I work on. I can do static type checking using [openapi-typescript](https://github.com/drwpow/openapi-typescript) beside runtime type checking. So we don't have to create validate functions to get static type checks like to auth package in Rafiki.

Issue #171, opened 2022-09-08
Enforce one access type per grant


wilsonianb:
https://github.com/interledger/open-payments/pull/168#discussion_r965248076

PR #181, opened 2022-09-16
chore: enforce one access type per grant

wilsonianb:
- fixes https://github.com/interledger/open-payments/issues/171

### DIFF
--- a/auth-server-open-api-spec.yaml
+++ b/auth-server-open-api-spec.yaml
@@ -550,7 +550,6 @@ components:
       discriminator:
         propertyName: type
         mapping:
-          account: '#/components/schemas/access-account'
           incoming-payment: '#/components/schemas/access-incoming'
           outgoing-payment: '#/components/schemas/access-outgoing'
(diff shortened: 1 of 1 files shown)


auth-server-open-api-spec.yaml
```diff
@@ -797,6 +796,25 @@ components:
           description: A description of the rights associated with this access token.
           items:
             $ref: '#/components/schemas/access'
+          allOf:
+            - contains:
+                properties:
+                  type:
+                    const: incoming-payment
```
sabineschaller, line 803: I tried to find the type `const` in the OpenAPI3.1 spec but I wasn't able to. Where did you find this?

wilsonianb: It's included in the version of JSON Schema used in OpenAPI 3.1
https://json-schema.org/understanding-json-schema/reference/generic.html#constant-values

sabineschaller (approved):

PR #183, opened 2022-09-16
chore: enforce base64 url-encoded sharedSecret

### LINKED ISSUE #175: Enforce sharedSecret as base64url

wilsonianb:
- fixes #175

### DIFF
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -103,7 +103,7 @@ paths:
                   value:
                     id: 'https://openpayments.guide/connections/016da9d5-c9a4-4c80-a354-86b915a04ff8'
                     ilpAddress: g.ilp.iwuyge987y.98y0f34tsrt8y
-                    sharedSecret: 6jR5iNIVRvqeasJeCty6C+YB5X9FhSOUPCL/5nha5Vs=
+                    sharedSecret: 1c7eaXa4rd2fFOBl1iydvCT1tV5TbM3RW1WLCafu_JA
                     assetCode: USD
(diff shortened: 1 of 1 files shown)


open-api-spec.yaml
```diff
@@ -838,7 +838,8 @@ components:
           readOnly: true
         sharedSecret:
           type: string
-          description: The shared secret to use when establishing a STREAM connection.
+          pattern: "^[a-zA-Z0-9-_]{43}$"
```
wilsonianb, line 841: The [STREAM spec](https://interledger.org/rfcs/0029-stream/#4-life-of-a-connection):
> Cryptographically secure random or pseudorandom shared secret (it is RECOMMENDED to use 32 bytes)

Should we not enforce the length here if 32 bytes technically isn't `REQUIRED`?
fwiw, ilp-pay has been enforcing 32-bytes:
https://github.com/interledgerjs/interledgerjs/blob/b3817e441f0c797cfd094bdbb5e4d6d046893613/packages/pay/src/open-payments.ts#L11
https://github.com/interledgerjs/interledgerjs/blob/b3817e441f0c797cfd094bdbb5e4d6d046893613/packages/pay/src/open-payments.ts#L276-L286

sabineschaller: I'm leaning towards not enforcing it if it's not required by the STREAM spec. ilp-pay is just an implementation that chooses to enforce it and our implementation can also enforce it. But I wouldn't enforce it in this spec.

sabineschaller (approved):

PR #188, opened 2022-09-21
chore(auth-spec): split auth spec

### LINKED ISSUE #123: Separate spec for RS<>AS API
Should the [`/introspect` endpoint](https://github.com/interledger/open-payments/blob/master/auth-server-open-api-spec.yaml#L349) be moved to its own OpenAPI spec, so that [client<>AS](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-3.2.1) and [RS<>AS](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-resource-servers) GNAP interactions are represented in distinct specs?

sabineschaller:
I split the spec into three. We may not want to have all of them in this repo. But for reference, they are all here for now.

- fixes #123


Client<>AS<>IdP-OAPI.yaml
```diff
@@ -0,0 +1,343 @@
+openapi: 3.1.0
```
wilsonianb, line 1: Should this file be named `AS<>IdP-OAPI.yaml`?

Client<>AS-OAPI.yaml
```diff
@@ -0,0 +1,698 @@
...
+      description: Management endpoint to revoke access token.
+      responses:
+        '204':
+          description: No Content
+        '401':
+          description: Unauthorized
+      tags:
+        - token
+components:
+  schemas:
```
wilsonianb, line 320: I think we should move the shared (or even all schemas) into a separate file (or individual files per schema?) and access them via relative remote references, such as:
```
$ref: 'schemas.json#/access
```
https://swagger.io/docs/specification/using-ref/

Client<>AS<>IdP-OAPI.yaml
wilsonianb: https://spec.openapis.org/oas/v3.1.0#document-structure
>It is RECOMMENDED that the root OpenAPI document be named: openapi.json or openapi.yaml.

Actually... what if these were in separate subdirectories?
Maybe something like (also changing the current [`open-api-spec.yaml`](https://github.com/interledger/open-payments/blob/main/open-api-spec.yaml))
```
openapi.yaml
AS/client/openapi.yaml
AS/IdP/openapi.yaml
AS/RS/openapi.yaml
```

sabineschaller: Even though this introduces a bunch of sub-directories, I like the structure.
Also, I love how you always dig up these things 👍

Client<>AS-OAPI.yaml
wilsonianb: maybe even `shared/schemas.json` if our current `open-api-spec.yaml` also ends up referencing it

Client<>AS<>IdP-OAPI.yaml
wilsonianb: Referencing relative yaml files across directories is starting to give me (codius) [kustomize](https://kustomize.io/) vibes
https://github.com/codius/codius-faas/blob/main/config/default/kustomization.yaml

wilsonianb (commented):
I just noticed workflows weren't run for this pr.
The openapi validator workflow should be updated with whatever new spec files we end up adding
https://github.com/interledger/open-payments/blob/main/.github/workflows/openapi.yaml

Client<>AS-OAPI.yaml
sabineschaller: I wonder, however, if readme.io can deal with shared schemas 🤔

.github/workflows/openapi.yaml
```diff
@@ -21,11 +31,30 @@ jobs:
...
 
-      - name: Validate open-api-spec.yaml
+      - name: AsyncAPI extension
         run: |
           echo "{\"extends\":[\"spectral:oas\",\"spectral:asyncapi\"]}" >> .spectral.json
+
+      # TODO: remove old ones
+      - name: Validate open-api-spec.yaml
+        run: |
           npx @stoplight/spectral-cli lint open-api-spec.yaml
```
wilsonianb, line 41:
```suggestion
          npx @stoplight/spectral-cli lint ./openapi-specs/**/openapi.yaml
```
https://meta.stoplight.io/docs/spectral/9ffa04e052cc1-spectral-cli

sabineschaller:
I think for the workflows to run they need to be on main already.

.github/workflows/openapi.yaml
sabineschaller: Are you suggesting doing that instead of having them all individually?

wilsonianb: Yeah

.github/workflows/openapi.yaml
```diff
@@ -5,14 +5,24 @@ on:
     branches:
       - main
     paths:
+      # TODO: remove old ones
       - auth-server-open-api-spec.yaml
       - open-api-spec.yaml
```
wilsonianb, line 10: Any reason not to go ahead and remove the old files?

.github/workflows/openapi.yaml
```diff
@@ -5,14 +5,24 @@ on:
...
+      - openapi-specs/openapi.yaml
+      - openapi-specs/AS/client/openapi.yaml
+      - openapi-specs/AS/IdP/openapi.yaml
+      - openapi-specs/AS/RS/openapi.yaml
```
wilsonianb, line 14:
```suggestion
      - 'openapi-specs/**'
```
ditto below

.github/workflows/openapi.yaml
sabineschaller: Not really. I kept them for reference. Deleted them now.

openapi-specs/openapi.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.1.0
 info:
   title: Open Payments
-  version: '1.0'
+  version: "1.0"
```
wilsonianb, line 4: Are these formatting changes intentional?
This seems to be the only file using double quotes.
I'm trying to see if we can make `spectral-cli` enforce formatting for us.
https://meta.stoplight.io/docs/spectral/01baf06bdd05a-rulesets
This just suggests using prettier (along with spectral):
https://github.com/Nexmo/api-specification#check-your-api-spec-for-validity-and-style

openapi-specs/openapi.yaml
```diff
@@ -1152,47 +1150,9 @@ components:
...
-          description: The asset code of the amount.
-        assetScale:
-          $ref: '#/components/schemas/assetScale'
-          description: The scale of the amount.
-      required:
-        - value
-        - assetCode
-        - assetScale
-      additionalProperties: false
     assetCode:
```
wilsonianb, line 1156: What do you think about also making `assetCode` and `assetScale` shared?

openapi-specs/openapi.yaml
sabineschaller: No, not intentional. I'll try to revert that.

sabineschaller: Enforcing would be great. But like you said, it may help to just include a `prettierrc.yml`.

wilsonianb: `shared/schemas.yaml` is also double quoted

openapi-specs/shared/schemas.yaml
```diff
@@ -0,0 +1,258 @@
+access:
```
wilsonianb, line 1:
```suggestion
access-item:
```
I think the access array type should be added to the shared schemas.
As is, the one access type entry per grant restriction is only defined in the RS<>Client spec:
https://github.com/interledger/open-payments/blob/23bd697583293884822c6aa9c0944ae8a13aea02/openapi-specs/AS/client/openapi.yaml#L437-L455

wilsonianb (commented):
How about:
`openapi-specs/openapi.yaml` --> `openapi-specs/RS/openapi.yaml`
or even
`openapi/RS/openapi.yaml`

openapi-specs/shared/schemas.yaml
sabineschaller: Good catch!

.prettierrc.yaml
```diff
@@ -0,0 +1 @@
+singleQuote: true
```
wilsonianb, line 1: Can we enforce this is the openapi GitHub Actions workflow?

wilsonianb (approved):

.prettierrc.yaml
sabineschaller: We could add [this](https://github.com/marketplace/actions/prettier-check) but there is no official Github Action for it.

.github/workflows/openapi.yaml
```diff
@@ -21,11 +19,10 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
 
-      - name: Validate open-api-spec.yaml
+      - name: AsyncAPI extension
```
wilsonianb, line 22:
```suggestion
      - name: Lint openapi.yaml
        run: |
          npx prettier --check openapi/**/openapi.yaml

      - name: AsyncAPI extension
```

.github/workflows/openapi.yaml
```diff
@@ -21,11 +19,10 @@ jobs:
...
         run: |
           echo "{\"extends\":[\"spectral:oas\",\"spectral:asyncapi\"]}" >> .spectral.json
-          npx @stoplight/spectral-cli lint open-api-spec.yaml
 
-      - name: Validate auth-server-open-api-spec.yaml
+      - name: Validate openapi.yaml
         run: |
-          npx @stoplight/spectral-cli lint auth-server-open-api-spec.yaml
+          npx @stoplight/spectral-cli lint openapi/../openapi.yaml
```
wilsonianb, line 28:
```suggestion
          npx @stoplight/spectral-cli lint openapi/**/openapi.yaml
```

adrianhopebailie:
@sabineschaller - we should pull everything except the `client <> AS` and `client <> RS` specs into Rafiki rather than have them here.

Only the APIs that the client consumes are part of the standard.

sabineschaller:
> @sabineschaller - we should pull everything except the client <> AS and client <> RS specs into Rafiki rather than have them here.

I'm happy to do that. I had everything here for reference but if we hare happy with the split, I can move the others out.

wilsonianb:
I'm _split_ on where RS <> AS belongs.
Token introspection is based on GNAP and schemas from client <> RS. It's not Rafiki specific.
Yes, Rafiki and other Open Payments implementation may allow for the same entity to operate the RS and AS, but it seems like anyone who doesn't SHOULD?/MUST? follow the RS <> AS spec.
I think Rafiki still is, aside from not requiring httpsig.

sabineschaller:
I agree with @wilsonianb. Even though AS<>RS is not the core spec, we modeled it after the GNAP resource server spec. So I'd keep it here, too.

.github/workflows/publish-openapi.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+      - uses: readmeio/rdme@7.2.0
+        name: Sync Open Payments API spec
+        with:
+          rdme: openapi openapi/RS/openapi.yaml --key=${{ secrets.README_API_KEY }} --id=${{ secrets.README_OP_API_ID }}
+
+      - uses: readmeio/rdme@7.2.0
+        name: Sync Auth Server API spec
+        with:
+          rdme: openapi openapi/AS/client/openapi.yaml --key=${{ secrets.README_API_KEY }} --id=${{ secrets.README_AUTH_API_ID }}
```
wilsonianb, line 30: Should AS <> RS be published?

wilsonianb (approved):

.github/workflows/publish-openapi.yaml
sabineschaller: We need to create the placeholder on readme.io first. Should I just merge it as is now and then we decide which ones to move out and which ones to publish to readme or should we decide before I merge?

adrianhopebailie:
The AS <> RS interface is not an interoperability requirement in the same way that the client <> X specs are.

It is convenient that it uses the same schemas etc but it's not required to do so.

The specs that are considered part of the Open Payments "standard" will have to be slow changing once they have been adopted because we have no control over implementations. We should avoid putting anything out as part of the "standard" until we have to.

Until there is a use case for an RS trusting a remote AS and they need a standard interface we should keep that interface out of the standard. If in future we find a need for a standard interface then we can propose the Rafiki interface as the standard.

.github/workflows/publish-openapi.yaml
wilsonianb: ship it :rocket: 
(plus I'm convinced by Adrian to remove AS <> RS, so it shouldn't be published)

PR #134, opened 2022-06-02
fix: stylistic tweaks for docs

adrianhopebailie:
- Consistent name for `id`
- OAS hack to provide descriptions on amount fields inside other schemas

PR #189, opened 2022-09-21
chore: add shared sub-resource id path parameter

wilsonianb:
We previously [renamed sub-resource ids](https://github.com/interledger/open-payments/pull/134) to `incomingPaymentId`, etc to disambiguate from the accompanying `accountId` path parameter.
`accountId` was removed in the [switch to payment pointers](https://github.com/interledger/open-payments/pull/145), and sub-resource ids are now the lone path parameters.
This makes path parameters consistent with the [auth server spec](https://github.com/interledger/open-payments/blob/6b5c088b71dd354139d5140c1a57aba6f7c9838e/auth-server-open-api-spec.yaml#L168).

- inspired by addition of reusable parameter definitions in #187

### DIFF
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -1304,7 +1304,6 @@ components:
       in: path
       schema:
         type: string
-        format: uuid
       description: Sub-resource identifier
       required: true
     signature:
--- a/open-api-spec.yaml
+++ b/open-api-spec.yaml
@@ -84,7 +84,7 @@ paths:
 
         The content should be slow changing and cacheable for long periods. Servers SHOULD use cache control headers.
       security: []
-  '/connections/{connectionId}':
+  '/connections/{id}':
     servers:
       - url: 'https://openpayments.guide/'
     get:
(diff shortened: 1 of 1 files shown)


open-api-spec.yaml
```diff
@@ -1328,6 +1299,14 @@ components:
     '403':
       description: Forbidden
   parameters:
+    id:
+      name: id
+      in: path
+      schema:
+        type: string
+        format: uuid
```
wilsonianb, line 1307: Yes or no on enforcing uuid?
https://github.com/interledger/rafiki/pull/603#discussion_r975017871

sabineschaller: I vote no. That's a business decision.

sabineschaller:
I like the change. I'm don't like that we need to change the implementation again. But I'm sure you'll do that @wilsonianb 😉

wilsonianb:
> I like the change. I'm don't like that we need to change the implementation again. But I'm sure you'll do that @wilsonianb wink

Yeah, I'll semi-revert this (as well as belatedly remove `accountId`):
https://github.com/interledger/rafiki/pull/373

sabineschaller (approved):

PR #190, opened 2022-09-23
feat: add JWKS endpoint

adrianhopebailie:
This adds a new path where the JWKs for a payment pointer can be fetched.


open-api-spec.yaml
```diff
@@ -84,6 +84,24 @@ paths:
 
         The content should be slow changing and cacheable for long periods. Servers SHOULD use cache control headers.
       security: []
+  /jwks.json:
```
wilsonianb, line 87: https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets
>Auth0 exposes a JWKS endpoint for each tenant, which is found at `https://YOUR_DOMAIN/.well-known/jwks.json`

That makes wonder if a payment pointer such as `https://alice.me/.well-known/pay` should have `https://alice.me/.well-known/jwk.json`

open-api-spec.yaml
```diff
@@ -813,6 +831,26 @@ components:
           assetCode: USD
           assetScale: 2
           authServer: 'https://openpayments.guide/auth'
+    jwks-document:
```
wilsonianb, line 834:
```suggestion
    json-web-key-set:
```

open-api-spec.yaml
adrianhopebailie: I'm not sure if I understand the question.

> That makes wonder if a payment pointer such as https://alice.me/.well-known/pay should have https://alice.me/.well-known/jwk.json

This proposal would resolve exactly like this. i.e. If you had a payment pointer `$alice.me` then your payment pointer URL would be `https://alice.me/.well-known/pay` and your keys would be at `https://alice.me/.well-known/jwk.json`

open-api-spec.yaml
```diff
@@ -84,6 +84,24 @@ paths:
...
+      summary: Get the keys bound to a Payment Pointer
+      tags:
+        - payment-pointer
+      responses:
+        '200':
+          description: JWKS Document Found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/jwks-document'
```
adrianhopebailie, line 98:
```suggestion
                $ref: '#/components/schemas/json-web-key-set'
```

open-api-spec.yaml
wilsonianb: As is, the payment pointer would be:
`https://alice.me/.well-known/pay`
and the keys would be at:
`https://alice.me/.well-known/pay/jwk.json`

I'm wondering if it's worth following the auth0 convention and putting the keys at:
`https://alice.me/.well-known/jwk.json`

(which would only affect `/.well-known/pay` path prefix payment pointers)

wilsonianb: >(which would only affect /.well-known/pay path prefix payment pointers)

An example of a non `/.well-known/pay` payment pointer:

`$wallet.example/alice`
`https://wallet.example/alice`

in which case it seems like the keys would have to be at:
`https://wallet.example/alice/jwk.json`

open-api-spec.yaml
```diff
@@ -813,6 +831,26 @@ components:
           assetCode: USD
           assetScale: 2
           authServer: 'https://openpayments.guide/auth'
+    json-web-key-set:
```
wilsonianb, line 834: Since this only includes keys and not client info, is the client (display) info expected to be found by querying the payment pointer itself?
Or is client display info something that the client provides in grant requests that the account provider doesn't attest to?
https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#section-2.3.2
For comparison, the current `/keys/{id}` endpoint in the Rafiki backend returns:
https://github.com/interledger/rafiki/blob/464ef810a8a47c718c4132a426c169c95d7ae8d7/packages/backend/src/clientKeys/routes.ts#L36-L45

adrianhopebailie: I think that pulling this data from the account provider is more reliable and opens us up to supporting use cases in future where this data can be verified (e.g. using DIDs and verifiable credentials)

adrianhopebailie: To be more specific, I think we should say that we only allow identifying the client by reference

wilsonianb: So how does the RS pull client display info from the account provider?

adrianhopebailie: What if we put this in the payment pointer response?

If you are a consumer that wants to protect your privacy then you wouldn't provide this but if you're a business and likely to act as a client then you would.

adrianhopebailie: Agree, but that won't be every time right?

wilsonianb: Yeah, it seems like the payment pointer (or AS grant) request would only happen once per grant?
Whereas the `jwks.json` request would be happening on every grant and token introspection request (unless the `auth` package starts caching keys).
It looks like the rafiki `auth` package is currently only storing `kid`. It seems like it should also store the public key (to bind to the grant)?
https://github.com/interledger/rafiki/blob/f17fdb9981d57d6199062a94ef032d47339b5459/packages/auth/migrations/20220504161932_create_grants_table.js#L15

adrianhopebailie: > the jwks.json request would be happening on every grant and token introspection request (unless the auth package starts caching keys)

I'd definitely suggest caching keys. We probably need to give some thought to an appropriate cache period.

wilsonianb: I think the request to the payment pointer to retrieve client display info will only need to happen on grant requests with `interact` specified.

open-api-spec.yaml
```diff
@@ -1280,6 +1318,44 @@ components:
...
+          type: string
+          enum:
+            - OKP
+        crv:
+          type: string
+          enum:
+            - Ed25519
+        x:
+          type: string
+          format: binary
```
wilsonianb, line 1340:
```suggestion
          pattern: "^[a-zA-Z0-9-_]+$"
          description: The base64 url-encoded public key.
```
https://www.rfc-editor.org/rfc/rfc7517#section-3

See `sharedSecret`:
https://github.com/interledger/open-payments/blob/4c873dba89164decffbe84905d12f1d4ec045389/open-api-spec.yaml#L832-L836

wilsonianb (approved):
~The file is in another castle~
Gotta the resolve the conflicts.

PR #204, opened 2022-10-21
chore: update shared schema spec for consistency

mkurapov:
There is an open [Rafiki PR](https://github.com/interledger/rafiki/pull/669) that generates Typescript types from the Open API spec using the [openapi-typescript](https://github.com/drwpow/openapi-typescript) package. 
Whenever type generation script runs, it generates a file with types as such:
```ts
export interface paths {
   ...
}

export interface components {
   schemas: {
      ...
   }
   responses: {
      ...
   }
   parameters: {
      ...
   }
}

export interface operations {
   ...
}

export interface external {
   ...
}
```

The `external` interface is the one that holds references to the types for when an external `$ref` is used in a spec (e.g. `$ref: ../shared/schemas.yaml#/amount`. 

However, after the spec split, the type generation for the resource server failed to populate this `external` interface, since `shared/schemas.yaml` was declaring components directly at the root like:
```yaml
assetCode:
  title: Asset code
  type: string
  description: The assetCode is a code that indicates the underlying asset. This SHOULD be an ISO4217 currency code.
```
The generator can only populate the external/shared schema interface if using the `components -> schemas` declaration, meaning the shared schema items must be declared like:

```yaml
components:
  schema:
    ...
    assetCode:
      title: Asset code
      type: string
      description: The assetCode is a code that indicates the underlying asset. This SHOULD be an ISO4217 currency code.
```

This PR nests the fields in `openapi/shared/schemas.yaml` to be under `components -> schemas`  in order to allow type generation to work. If this is too much of a change to support this tool, I might be able to figure out a different (possibly convoluted) solution, but I thought this standardization wouldn't cause issues nor be too much of a stretch.

### DIFF
--- a/openapi/shared/schemas.yaml
+++ b/openapi/shared/schemas.yaml
@@ -1,282 +1,295 @@
-access:
-  type: array
-  description: A description of the rights associated with this access token.
-  items:
-    $ref: '#/access-item'
-  allOf:
-    - contains:
-        properties:
-          type:
-            const: incoming-payment
-      minContains: 0
-      maxContains: 1
-    - contains:
-        properties:
-          type:
-            const: outgoing-payment
-      minContains: 0
-      maxContains: 1
-    - contains:
-        properties:
-          type:
-            const: quote
-      minContains: 0
-      maxContains: 1
-access-item:
-  discriminator:
-    propertyName: type
-    mapping:
-      incoming-payment: '#/access-incoming'
-      outgoing-payment: '#/access-outgoing'
-      quote: '#/access-quote'
-  oneOf:
-    - $ref: '#/access-incoming'
-    - $ref: '#/access-outgoing'
-    - $ref: '#/access-quote'
-  description: The access associated with the access token is described using objects that each contain multiple dimensions of access.
-  unevaluatedProperties: false
-access-incoming:
-  title: access-incoming
-  type: object
-  properties:
-    type:
-      type: string
-      enum:
-        - incoming-payment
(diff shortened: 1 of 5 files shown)


openapi/shared/schemas.yaml
```diff
@@ -1,282 +1,295 @@
...
+info:
+  title: Open Payments - Shared schemas
+  version: '1.0'
+  license:
+    name: Apache 2.0
+    identifier: Apache-2.0
+  summary: Open Payments - Shared schemas
+  description: 'Shared schemas used across Open Payments APIs'
+  contact:
+    email: tech@interledger.org
```
mkurapov, line 11: not necessary perse, but just additional information

openapi/shared/schemas.yaml
```diff
@@ -1,282 +1,295 @@
...
+components:
+  schemas:
```
mkurapov, line 13: no changes other than this nesting

wilsonianb (approved):
related issue:
- https://github.com/drwpow/openapi-typescript/issues/892

We can consider reverting this if/when that's resolved.

mkurapov:
@wilsonianb yes, that's the one I was looking at, the [final response](https://github.com/drwpow/openapi-typescript/issues/892#issuecomment-1224308930) in that issue was a bit different than the [earlier one](https://github.com/drwpow/openapi-typescript/issues/892#issuecomment-1177786341) in that thread that was relevant to this change. I hope if there is a change to the library both of those fixes will be addressed.

PR #205, opened 2022-10-21
Move Rafiki specs to Rafiki

adrianhopebailie:
This PR is linked to a PR that is pending on Rafiki
It moves the IdP and RS specs for the AS into the Rafiki repo


openapi/auth-server.yaml
```diff
@@ -437,3 +437,5 @@ components:
       name: Authorization
       type: apiKey
       in: header
+security:
+  - GNAP: []
```
wilsonianb, line 441: Does this belong?
```suggestion
```

adrianhopebailie: Yes, it fails validation without it

wilsonianb (approved):

PR #208, opened 2022-11-03
feat: use nullable limits and openapi-compliant uniqueness for access schema

njlie:
Addresses (but doesn't really fix) https://github.com/interledger/rafiki/issues/630. This spec is less strict to allow valid responses from the AS to pass validation, but is a bit too loose for what we actually want to achieve with it. The main restriction that this doesn't cover is:
* The spec does not restrict a grant to (at most) one of each `type` of access.

### DIFF
--- a/openapi/schemas.yaml
+++ b/openapi/schemas.yaml
@@ -16,25 +16,8 @@ components:
       description: A description of the rights associated with this access token.
       items:
         $ref: '#/components/schemas/access-item'
-      allOf:
-        - contains:
-            properties:
-              type:
-                const: incoming-payment
-          minContains: 0
-          maxContains: 1
-        - contains:
-            properties:
-              type:
-                const: outgoing-payment
-          minContains: 0
-          maxContains: 1
-        - contains:
-            properties:
-              type:
-                const: quote
-          minContains: 0
-          maxContains: 1
+      uniqueItems: true
+      maxItems: 3
     access-item:
       discriminator:
         propertyName: type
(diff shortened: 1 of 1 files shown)


openapi/schemas.yaml
```diff
@@ -259,6 +242,7 @@ components:
       title: limits-outgoing
       description: Open Payments specific property that defines the limits under which outgoing payments can be created.
       type: object
+      nullable: true
```
wilsonianb, line 245: I'm struggling to find `nullable` in OpenAPI or JSON Schema docs

njlie: https://swagger.io/docs/specification/data-models/data-types/#null

wilsonianb: It looks like `nullable` was removed in OpenAPI 3.1
https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0

openapi/schemas.yaml
```diff
@@ -258,7 +241,9 @@ components:
     limits-outgoing:
       title: limits-outgoing
       description: Open Payments specific property that defines the limits under which outgoing payments can be created.
-      type: object
+      type:
+        - object
+        - null
```
wilsonianb, line 246: Based on examples, I'm seeing at least `null` (and maybe `object`) might need to be a "value"
```suggestion
        - 'null'
```
https://github.com/OAI/OpenAPI-Specification/issues/2244#issuecomment-899084964

Not sure if that will resolve the validator errors

wilsonianb (approved):

PR #216, opened 2022-11-10
fix: move workflow_run in publish action

wilsonianb:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

### DIFF
--- a/.github/workflows/publish-openapi.yaml
+++ b/.github/workflows/publish-openapi.yaml
@@ -1,19 +1,14 @@
 name: OpenAPI Publisher
 
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - openapi/**
+  workflow_run:
+    workflows: ['OpenAPI Validator']
+    branches: [main]
+    types:
+      - completed
 
 jobs:
   publish:
-    workflow_run:
-      workflows: ['OpenAPI Validator']
-      branches: [main]
-      types:
-        - completed
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-openapi.yaml
```diff
@@ -1,19 +1,14 @@
...
-  push:
-    branches:
-      - main
-    paths:
-      - openapi/**
+  workflow_run:
+    workflows: ['OpenAPI Validator']
+    branches: [main]
+    types:
+      - completed
```
mkurapov, line 8: do we need the `types` here?

wilsonianb: >By default, all activity types trigger workflows that run on this event. 

I think excluding would cause the publisher to run without the validator necessarily succeeding (via `requested` or `in_progress` types)

mkurapov (approved):

.github/workflows/publish-openapi.yaml
mkurapov: Would it be good to add an `on-success` flag?
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow

mkurapov: Otherwise, if I'm understanding the docs correctly, we would still publish if the validator failed. honestly would have made sense to have a `type: succeeded` but 🤷

PR #226, opened 2022-12-06
chore: move json-web-key to shared schemas

wilsonianb:
To replace:
https://github.com/interledger/rafiki/blob/a0302b390192f001fc4baa6bc55ae8dcfb13143a/packages/auth/src/openapi/resource-server.yaml#L117-L170

### DIFF
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -838,7 +838,7 @@ components:
         keys:
           type: array
           items:
-            $ref: '#/components/schemas/json-web-key'
+            $ref: ./schemas.yaml#/components/schemas/json-web-key
           readOnly: true
       required:
         - keys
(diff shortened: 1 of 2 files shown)


openapi/resource-server.yaml
```diff
@@ -838,7 +838,7 @@ components:
         keys:
           type: array
           items:
-            $ref: '#/components/schemas/json-web-key'
+            $ref: ./schemas.yaml#/components/schemas/json-web-key
```
mkurapov, line 841: will this be shared elsewhere? In the auth-server schema?

mkurapov (approved):

openapi/resource-server.yaml
wilsonianb: I'm planning to replace this
https://github.com/interledger/rafiki/blob/a0302b390192f001fc4baa6bc55ae8dcfb13143a/packages/auth/src/openapi/resource-server.yaml#L117-L170
with
```
$ref: ./schemas.yaml#/components/schemas/json-web-key`
```
:thinking: :thinking: :thinking: :thinking: 
but now that I think about it, that spec file won't have a key at all after:
- https://github.com/interledger/rafiki/issues/737

wilsonianb: :point_up: that happened, so I'm closing this

PR #228, opened 2022-12-07
feat: add form style to pagination params

### LINKED ISSUE #227: Update pagination param in the API guide
The Open Payments API doc examples prepend the `pagination` parameter when showing an example call to list outgoing or incoming payments:

Screen Shot 2022-12-07 at 2 56 31 PM

Specifically,

```sh
curl --request GET \
     --url 'https://openpayments.guide/alice/incoming-payments?pagination=first=10&cursor=uuid' \
     --header 'accept: application/json'
```

https://docs.openpayments.guide/reference/list-outgoing-payments

However, we do not use root `pagination` parameter when we parse the pagination query parameters when we are handling those calls in `backend`, we only look at `first` `last` `cursor` params:

https://github.com/interledger/rafiki/blob/a0302b390192f001fc4baa6bc55ae8dcfb13143a/packages/backend/src/open_payments/payment_pointer/routes.ts#L50

https://github.com/interledger/rafiki/blob/a0302b390192f001fc4baa6bc55ae8dcfb13143a/packages/backend/src/shared/pagination.ts#L9-L20

We should update Open Payments API docs to ignore `pagination` as the root key parameter for the example requests.

ie, it should be just
```sh
curl --request GET \
     --url 'https://openpayments.guide/alice/incoming-payments?first=10&cursor=uuid' \
     --header 'accept: application/json'
```

mkurapov:
Fixes #227 
_Copied from #227 description:_

The Open Payments API doc examples prepend the `pagination` parameter when showing an example call to list outgoing or incoming payments:

Screen Shot 2022-12-07 at 2 56 31 PM

Specifically,

```sh
curl --request GET \
     --url 'https://openpayments.guide/alice/incoming-payments?pagination=first=10&cursor=uuid' \
     --header 'accept: application/json'
     
// inclusion of "pagination" is the issue here    
```

https://docs.openpayments.guide/reference/list-outgoing-payments

However, we do not use root `pagination` parameter when we parse the query parameters when we are handling `GET /(incoming | outgoing)-payments` calls in `backend`, we only look at `first` `last` `cursor` params:

https://github.com/interledger/rafiki/blob/a0302b390192f001fc4baa6bc55ae8dcfb13143a/packages/backend/src/open_payments/payment_pointer/routes.ts#L50

https://github.com/interledger/rafiki/blob/a0302b390192f001fc4baa6bc55ae8dcfb13143a/packages/backend/src/shared/pagination.ts#L9-L20

We should update Open Payments API docs to ignore `pagination` as the root key parameter for the example requests.

ie, it should be just
```sh
curl --request GET \
     --url 'https://openpayments.guide/alice/incoming-payments?first=10&cursor=uuid' \
     --header 'accept: application/json'
```

### DIFF
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -517,11 +517,7 @@ paths:
           $ref: '#/components/responses/403'
       description: List all outgoing payments on the payment pointer
       parameters:
-        - schema:
-            $ref: '#/components/schemas/pagination'
-          in: query
-          name: pagination
-          description: Pagination parameters
+        - $ref: '#/components/parameters/pagination'
         - $ref: '#/components/parameters/signature-input'
         - $ref: '#/components/parameters/signature'
       tags:
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -324,6 +324,7 @@ paths:
           in: query
           name: pagination
           description: Pagination parameters
+          style: form
         - $ref: '#/components/parameters/signature-input'
         - $ref: '#/components/parameters/signature'
       tags:
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -319,12 +319,7 @@ paths:
           $ref: '#/components/responses/403'
       description: List all incoming payments on the payment pointer
       parameters:
-        - schema:
-            $ref: '#/components/schemas/pagination'
-          in: query
-          name: pagination
-          description: Pagination parameters
-          style: form
+        - $ref: '#/components/parameters/pagination'
         - $ref: '#/components/parameters/signature-input'
         - $ref: '#/components/parameters/signature'
       tags:
@@ -1322,6 +1317,13 @@ components:
     '403':
       description: Forbidden
   parameters:
+    pagination:
+      schema:
+        $ref: '#/components/schemas/pagination'
+      in: query
+      name: pagination
+      description: Pagination parameters
+      style: form
     id:
       name: id
       in: path


openapi/resource-server.yaml
```diff
@@ -1321,6 +1313,13 @@ components:
     '403':
       description: Forbidden
   parameters:
+    pagination:
+      schema:
+        $ref: '#/components/schemas/pagination'
+      in: query
+      name: pagination
+      description: Pagination parameters
+      style: form
```
mkurapov, line 1322: `style: form` is the fix here

mkurapov:
Tested on `integration` branch:
Screen Shot 2022-12-07 at 3 21 25 PM

openapi/resource-server.yaml
wilsonianb: :confused: 
It sounds like `form` is the default with `in: query`
https://spec.openapis.org/oas/v3.1.0#fixed-fields-9

>Describes how the parameter value will be serialized depending on the type of the parameter value. Default values (based on value of `in`): for `query` - `form`

Does that indicate there's an issue with readme.io's OpenAPI support?

wilsonianb (approved):

openapi/resource-server.yaml
mkurapov: Maybe because of the fact that the parameter was referencing a schema in `components/schemas/pagination` directly with `name: pagination` it defaulted to `style: deepObject`?

Issue #136, opened 2022-06-03
Update pagination parameters and response


wilsonianb:
Discussion:
https://github.com/interledger/open-payments/pull/129#discussion_r885805243

PR #241, opened 2023-02-07
don't define pagination query parameters as object

wilsonianb:

- Don't define pagination query parameters as object
  Existing OpenAPI 3.1 Node.js tooling cannot explode to individual parameters.
  Enforce mutually exclusive validation, etc in the RS.

## Context

- https://github.com/interledger/rafiki/issues/443

### DIFF
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -319,7 +319,9 @@ paths:
           $ref: '#/components/responses/403'
       description: List all incoming payments on the payment pointer
       parameters:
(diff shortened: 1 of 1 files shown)


wilsonianb:
I'm still open to :point_down: but decided to stick with the existing parameters to limit the scope here
- https://github.com/interledger/open-payments/issues/136

sabineschaller (approved):

mkurapov (approved):

PR #242, opened 2023-02-22
docs: move info from readme into spec

sabineschaller:

We currently have descriptions for incoming and outgoing payments on 
Readme.io which are outdated. I pulled in all the info that is still valid such that we can remove the outdated docs from Readme.io

### DIFF
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -14,7 +14,7 @@ info:
 
     An outgoing payment is an instruction to make a payment out of the payment pointer.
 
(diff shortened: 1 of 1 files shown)


wilsonianb:
The descriptions live in readme and not the repo?

sabineschaller:
The only two that *had* descriptions on Readme were
- create incoming payment
- create outgoing payment

I think @adrianhopebailie already deleted it. It had information about state transitions. E.g. outgoing payment is created in a `pending` state, then moves to `processing`, then `completed` or `failed`...

wilsonianb (approved):

PR #266, opened 2023-06-06
chore: make startCursor and endCursor optional

### LINKED ISSUE #191: Make `startCursor`/`endCursor` optional
The `pageInfo` object in `list` responses should not require `startCursor`/`endCursor`
https://github.com/interledger/open-payments/blob/4c873dba89164decffbe84905d12f1d4ec045389/open-api-spec.yaml#L1278-L1279

They are technically required in the [relay spec](https://relay.dev/graphql/connections.htm#sec-undefined.PageInfo.Fields), but there is an [open issue](https://github.com/graphql/graphql-relay-js/issues/243) about making them optional (for when the result is empty), and they are currently optional in Rafiki:
- https://github.com/interledger/rafiki/blob/3d30ce755b4218a777ec77f408be5d15dfa0e687/packages/backend/src/shared/pagination.ts#L26-L30
- https://github.com/interledger/rafiki/blob/3d30ce755b4218a777ec77f408be5d15dfa0e687/packages/backend/src/graphql/schema.graphql#L134-L143

We could forgo this change by instead not having the Open Payments API list responses match relay:
- #136

mkurapov:

- Making `startCursor` and `endCursor` optional in the resource server spec for pagination responses.
- Adding CI check for missing generated types in the open-payments client

## Context

- Fixes #191 

When there are no results, it makes sense for `startCursor` and `endCursor` to be nullable:
- https://github.com/interledger/open-payments/issues/191#issuecomment-1579003465

This is actually causing an error in Rafiki when making Open Payments requests to list incoming/outgoing payments, and there are no results.

See the new changes here: https://open-payments-integration.readme.io/v1.1/reference/list-incoming-payments

### DIFF
--- a/.changeset/metal-panthers-fix.md
+++ b/.changeset/metal-panthers-fix.md
@@ -1,0 +1,5 @@
+---
+'@interledger/open-payments': minor
+---
+
+Optional startCursor & endCursor for pagination
--- a/.github/workflows/lint-test-build.yaml
+++ b/.github/workflows/lint-test-build.yaml
@@ -41,7 +41,7 @@ jobs:
         id: verify-changed-files
         with:
           files: |
-            **/generated/*.ts
+            packages/open-payments/**/generated/*.ts
       - name: Fail if missing new open-payments types
         if: steps.verify-changed-files.outputs.files_changed == 'true'
         run: exit 1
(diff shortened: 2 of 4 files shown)


mkurapov:
@sabineschaller I'd like to use this new spec in Rafiki, should I bump the version and use the updated one in Rafiki?

BlairCurrey (approved):

BlairCurrey (approved):

Issue #221, opened 2022-11-17
Deprecate `description` and `externalRef` in favor of `metadata`


sabineschaller:
In some use cases, `description` and `externalRef` are overkill, where in others, it may not be enough. I vote for using `metadata` instead. Making it type `string` and stringify everything would probably be easiest.

wilsonianb:
Or memos :stuck_out_tongue_winking_eye: 
https://xrpl.org/transaction-common-fields.html#memos-field

adrianhopebailie:
Memos seems like a nice way to have SOME structure.
Basically an array of objects with a rel, type, and value.
(Odd that memos in XRPL use `format` for the `mime-type`)

Format can default to `text/plain`

```json
{
  "memos": [
    {
      "rel": "invoiceNumber",
      "value": "INV542654"
    },
    {
      "rel": "description",
      "value": "Payment for the widgets"
    },
    {
      "rel": "invoice",
      "type": "application/pdf"
      "value": "<base64encoded PDF>"
    } 
  ]
}

PR #267, opened 2023-06-08
feat: add  metadata field to payments

BlairCurrey:
- adds `metadata` field to incoming/outgoing payment routes

## Context

First of several PRs for #221 (add metadata to spec, implement metadata field in rafiki, remove `description`/`externalRef`)

### DIFF
--- a/.changeset/khaki-spies-tan.md
+++ b/.changeset/khaki-spies-tan.md
@@ -1,6 +1,5 @@
 ---
 '@interledger/open-payments': minor
-'@interledger/openapi': minor
 ---
 
 Added metadata field to payments
--- a/.changeset/khaki-spies-tan.md
+++ b/.changeset/khaki-spies-tan.md
@@ -1,0 +1,6 @@
+---
+'@interledger/open-payments': minor
+'@interledger/openapi': minor
+---
+
+Added metadata field to payments
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -168,7 +168,7 @@ paths:
                     completed: false
                     expiresAt: '2022-02-03T23:20:50.52Z'
                     externalRef: INV2022-02-0137
-                    metadata: '{"externalRef": "INV2022-02-0137"}'
+                    metadata: { externalRef: 'INV2022-02-0137' }
                     ilpStreamConnection:
                       id: 'https://openpayments.guide/connections/ff394f02-7b7b-45e2-b645-51d04e7c345c'
                       ilpAddress: g.ilp.iwuyge987y.98y08y
@@ -214,7 +214,7 @@ paths:
                     assetCode: USD
                     assetScale: 2
                   externalRef: INV2022-02-0137
-                  metadata: '{"externalRef": "INV2022-02-0137"}'
+                  metadata: { externalRef: 'INV2022-02-0137' }
         description: |-
           A subset of the incoming payments schema is accepted as input to create a new incoming payment.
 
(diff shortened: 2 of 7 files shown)


BlairCurrey:
I was going to add a case in this test https://github.com/interledger/open-payments/blob/0fea2947d75334216eae9cc21599cf032c24954a/packages/openapi/src/middleware.test.ts#L159 but this test file uses a hard-coded link to the openapi spec on rafiki, which means it wont know about the new metadata field. https://github.com/interledger/open-payments/blob/0fea2947d75334216eae9cc21599cf032c24954a/packages/openapi/src/middleware.test.ts#L38-L39 

I will need to follow up with another PR to add the test case and probably change the hardcoded link once this one is merged, or maybe I should just change to a local path (depending on environment, so it still works in ci)?

openapi/resource-server.yaml
```diff
@@ -1133,6 +1161,10 @@ components:
         externalRef:
           type: string
           description: A reference that can be used by external systems to reconcile this payment with their systems. E.g. An invoice number. (Optional)
+        metadata:
+          type: string
+          format: json
+          description: Additional metadata associated with the incoming payment, represented as a JSON string. (Optional)
```
mkurapov, line 1167:
```suggestion
          description: Additional metadata associated with the outgoing payment, represented as a JSON string. (Optional)
```

.changeset/khaki-spies-tan.md
```diff
@@ -0,0 +1,6 @@
+---
+'@interledger/open-payments': minor
+'@interledger/openapi': minor
```
mkurapov, line 3: I don't think we need to change the `package/openapi`, unless I'm missing something

openapi/resource-server.yaml
```diff
@@ -200,6 +201,10 @@ paths:
                 externalRef:
                   type: string
                   description: A reference that can be used by external systems to reconcile this payment with their systems. E.g. An invoice number. (Optional)
+                metadata:
+                  type: string
+                  format: json
+                  description: Additional metadata associated with the incoming payment, represented as a JSON string. (Optional)
```
mkurapov, line 207: Is `format: json` valid? Not sure if we need this if `type: string`.

If we want to limit it to an object with free-form fields, we could also do this instead:
```yaml
    metadata:
      type: object
      additionalProperties: true
```

We could discuss in slack with Sabine as well

mkurapov, line 207:
```suggestion
                metadata:
                  type: string
                  format: json
                  description: Additional metadata associated with the incoming payment, represented as a string. (Optional)
```
We can have the JSON string in (some?all?) of the examples, but any string is valid

BlairCurrey: I double-checked and json is not a valid format. I will remove if not going the object route.

In graphql we need to use a string (no "any" type input). That's mostly what informed my direction here but I suppose if we want to enforce the metadata as an object then the object type would be better.

I think it probably is a good idea that we ensure the metadata is a valid object. That was the original use-case and I think it would just make more sense when we get incoming payments (no logic to determine what metadata is).

BlairCurrey: That is a good point. I was assuming it should be an object so I was including that note to show that intention since the route will accept any string. Guess I should remove "represented as a JSON string" regardless of what we decide about keeping as string vs. changing to object.

.changeset/khaki-spies-tan.md
BlairCurrey: No I think you're right. I was thinking it was the `./openapi/resource-server.yaml` schema changes but that is not the package. It showed up as a changed package when running `pnpm changeset:add` but now I'm realizing that was from a local, uncommitted change. Will remove.

openapi/resource-server.yaml
BlairCurrey: And on the topic of graphql, I guess an arbitrary object is possible with something like [graphql-type-json](https://www.npmjs.com/package/graphql-type-json)

BlairCurrey: switching to an object here and will implement it as a json scalar in gql using `graphql-scalars` (a package we already use)

mkurapov (approved):
Looks good, tested out on here https://open-payments-integration.readme.io/v1.1/reference/create-incoming-payment


Also, I think having metadata as he object is good, it is what Stripe does as well: https://stripe.com/docs/api/metadata

mkurapov (approved):

PR #270, opened 2023-06-29
chore: remove deprecated payment fields

BlairCurrey:

- Removes `externalRef` and `description` fields from payments (breaking change)
- Major version bump of `open-payments`

## Context

Fixes: #221

### DIFF
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -283,7 +283,10 @@ paths:
                         expiresAt: '2022-04-12T23:20:50.52Z'
                         createdAt: '2022-03-12T23:20:50.52Z'
                         updatedAt: '2022-04-01T10:24:36.11Z'
-                        description: 'I love your website, Alice! Thanks for the great content'
+                        metadata:
+                          {
+                            description: 'I love your website, Alice! Thanks for the great content'
+                          }
                         completed: false
                 backward pagination:
                   value:
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -376,7 +376,7 @@ paths:
                       value: '0'
                       assetCode: USD
                       assetScale: 2
-                      metadata: { description: Thank you for the shoes. }
+                    metadata: { description: Thank you for the shoes. }
                     createdAt: '2022-03-12T23:20:50.52Z'
                     updatedAt: '2022-04-01T10:24:36.11Z'
         '401':
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -1,7 +1,7 @@
 openapi: 3.1.0
 info:
   title: Open Payments
-  version: '1.0'
+  version: '1.1'
   license:
     name: Apache 2.0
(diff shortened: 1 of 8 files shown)


packages/openapi/src/middleware.test.ts
```diff
@@ -155,8 +155,6 @@ describe('OpenAPI Validator', (): void => {
       ${{ incomingAmount: { value: '-2', assetCode: 'USD', assetScale: 2 } }} | ${'body.incomingAmount.value must match format "uint64"'}                       | ${'invalid incomingAmount, value non-positive'}
       ${{ incomingAmount: { value: '2', assetCode: 4, assetScale: 2 } }}      | ${'body.incomingAmount.assetCode must be string'}                               | ${'invalid incomingAmount, assetCode not string'}
       ${{ incomingAmount: { value: '2', assetCode: 'USD', assetScale: -2 } }} | ${'body.incomingAmount.assetScale must be >= 0'}                                | ${'invalid incomingAmount, assetScale negative'}
-      ${{ description: 123 }}                                                 | ${'body.description must be string'}                                            | ${'invalid description'}
-      ${{ externalRef: 123 }}                                                 | ${'body.externalRef must be string'}                                            | ${'invalid externalRef'}
```
BlairCurrey, line 159: I tried to add `metadata` here but this tests uses an old version of the spec (hardcoded on line 39): https://github.com/interledger/open-payments/raw/3930448672cfc678ec2bc02938566a316d83871c/open-api-spec.yaml

I tried to replace with https://github.com/interledger/open-payments/raw/97ad3e99265c3e1eb516eb678178b9f6a4e5065e/openapi/resource-server.yaml but all tests failed 🤷‍♂️

openapi/resource-server.yaml
```diff
@@ -384,7 +376,7 @@ paths:
                       value: '0'
                       assetCode: USD
                       assetScale: 2
-                    description: Thank you for the shoes.
+                      metadata: { description: Thank you for the shoes. }
```
mkurapov, line 379:
```suggestion
                    metadata: { description: Thank you for the shoes. }
```

openapi/resource-server.yaml
```diff
@@ -167,7 +167,6 @@ paths:
                       assetScale: 2
```
mkurapov, line 167: should we bump the version of the spec? 
There's a few different ways to do this, Stripe has an interesting approach where they use dates instead of version numbers when breaking changes are released:
- https://stripe.com/docs/api/versioning
- https://news.ycombinator.com/item?id=13708927
- https://keygen.sh/blog/breaking-things-without-breaking-things/

mkurapov: Probably should be a wider discussion, but I think we do need to version this change somehow

BlairCurrey: Yes, definitely agree since this is a breaking change. I'm reading those sources and some others and think we should talk about this further.

mkurapov (approved):

PR #275, opened 2023-07-31
feat: add properties for quoting

### LINKED ISSUE #274: New quoting mechanism

sabineschaller:

- updates resource server spec and auth server spec to change `sendAmount` to `debitAmount`

## Context

We originally planned to add fees to the OP resources, however, we decided that fees should be included in the `debitAmount`. The difference in `debitAmount - receiveAmount` can be considered being the fees. The ASE may list those fees separately on the consent screen displayed to the user. 

- fixes #274

### DIFF
--- a/.changeset/tender-ravens-breathe.md
+++ b/.changeset/tender-ravens-breathe.md
@@ -1,0 +1,5 @@
+---
+'@interledger/open-payments': minor
+---
+
+Adding properties for new quoting mechanism
--- a/.changeset/tender-ravens-breathe.md
+++ b/.changeset/tender-ravens-breathe.md
@@ -1,5 +1,5 @@
 ---
-'@interledger/open-payments': minor
+'@interledger/open-payments': major
 ---
 
 Adding properties for new quoting mechanism
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -199,9 +199,6 @@ paths:
                 metadata:
                   type: object
                   description: Additional metadata associated with the incoming payment. (Optional)
-                feeStructure:
-                  type: boolean
-                  description: Request receiving fee structure
             examples:
               Create incoming payment for $25 to pay invoice INV2022-02-0137:
                 value:
@@ -211,7 +208,6 @@ paths:
                     assetScale: 2
                   metadata:
                     externalRef: INV2022-02-0137
-                  feeStructure: true
         description: |-
           A subset of the incoming payments schema is accepted as input to create a new incoming payment.
 
@@ -361,7 +357,7 @@ paths:
                     paymentPointer: 'https://openpayments.guide/alice/'
                     failed: false
                     receiver: 'https://openpayments.guide/bob/incoming-payments/48884225-b393-4872-90de-1b737e2491c2'
-                    maxSendAmount:
+                    debitAmount:
                       value: '2600'
                       assetCode: USD
                       assetScale: 2
@@ -373,15 +369,6 @@ paths:
                       value: '0'
                       assetCode: USD
                       assetScale: 2
-                    fees:
-                      sendFeeAmount:
-                        value: '100'
-                        assetCode: USD
-                        assetScale: 2
-                      receiveFeeAmount:
-                        value: '0'
-                        assetCode: USD
-                        assetScale: 2
                     metadata:
                       description: Thank you for the shoes.
                     createdAt: '2022-03-12T23:20:50.52Z'
@@ -460,7 +447,7 @@ paths:
                           value: '2500'
                           assetCode: USD
                           assetScale: 2
-                        maxSendAmount:
+                        debitAmount:
                           value: '2600'
                           assetCode: USD
                           assetScale: 2
@@ -468,15 +455,6 @@ paths:
                           value: '2500'
                           assetCode: USD
                           assetScale: 2
-                        fees:
-                          sendFeeAmount:
-                            value: '100'
-                            assetCode: USD
-                            assetScale: 2
-                          receiveFeeAmount:
-                            value: '0'
-                            assetCode: USD
-                            assetScale: 2
                         createdAt: '2022-03-12T23:20:50.52Z'
                         updatedAt: '2022-04-01T10:24:36.11Z'
                         metadata:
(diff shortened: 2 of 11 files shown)


openapi/resource-server.yaml
```diff
@@ -1324,6 +1424,27 @@ components:
           kty: OKP
           crv: Ed25519
           x: oy0L_vTygNE4IogRyn_F5GmHXdqYVjIXkWs2jky7zsI
+    feeStructure:
+      title: feeStructure
+      type: object
+      properties:
+        fixed:
+          $ref: ./schemas.yaml#/components/schemas/amount
```
mkurapov, line 1432: Not directly related to the PR, but the description for the amount schema seems odd:

"All amounts are maxima, i.e. multiple payments can be created under a grant as long as the total amounts of these payments do not exceed the maximum amount per interval as specified in the grant."

openapi/resource-server.yaml
```diff
@@ -1324,6 +1424,27 @@ components:
...
+        percentage:
+          type: string
+    fees:
```
mkurapov, line 1435: additionalProperties: false?

openapi/resource-server.yaml
```diff
@@ -1324,6 +1424,27 @@ components:
           kty: OKP
           crv: Ed25519
           x: oy0L_vTygNE4IogRyn_F5GmHXdqYVjIXkWs2jky7zsI
+    feeStructure:
+      title: feeStructure
+      type: object
+      properties:
+        fixed:
+          $ref: ./schemas.yaml#/components/schemas/amount
+        percentage:
```
mkurapov, line 1433: should we specify what percentage is, ie. whether 1% should be specified as 0.01 vs 1?
Alternatively, we can just specify an additional example that includes the feeStructure

openapi/resource-server.yaml
```diff
@@ -1324,6 +1424,27 @@ components:
...
+          type: string
+    fees:
+      title: fees
+      type: object
+      properties:
+        sendFeeAmount:
+          $ref: ./schemas.yaml#/components/schemas/amount
+          description: Sending Accoung Servicing Entity's fees + ILP network fees
```
mkurapov, line 1441:
```suggestion
          description: Sending Accoung Servicing Entity's fees + possible ILP network fees
```
maybe?

openapi/resource-server.yaml
```diff
@@ -185,28 +186,32 @@ paths:
...
                   format: date-time
                   writeOnly: true
                 metadata:
                   type: object
-                  additionalProperties: true
                   description: Additional metadata associated with the incoming payment. (Optional)
-              additionalProperties: false
+                feeStructure:
+                  type: boolean
+                  description: Request receiving fee structure
```
mkurapov, line 204:
```suggestion
                  description: Request receiving fee structure (if sender required to pay receiving fees)
```
Maybe?

openapi/resource-server.yaml
```diff
@@ -576,14 +612,23 @@ paths:
                     id: 'https://openpayments.guide/alice/quotes/8c68d3cc-0a0f-4216-98b4-4fa44a6c88cf'
```
mkurapov, line 612: should the quote request body still include `sendAmount` or be changed to `maxSendAmount`?
Even though it doesn't match `maxSendAmount` I _think_ `sendAmount` is good to leave as is

.changeset/tender-ravens-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@interledger/open-payments': minor
```
mkurapov, line 2: I think this should be major since it contains breaking changes

openapi/resource-server.yaml
sabineschaller: I guess we didn't update it properly when moving it to the shared schema. I'll fix that in this PR as well.

openapi/resource-server.yaml
sabineschaller: I talked to @BlairCurrey and we were thinking about using basis points and integers instead. What do you think?

.changeset/tender-ravens-breathe.md
sabineschaller: I must have run the wrong command 🙃

openapi/resource-server.yaml
BlairCurrey: I was thinking about this as well while working on https://github.com/interledger/rafiki/issues/1639. I think it might be confusing to have `sendAmount` going in an `maxSendAmount` coming back. The user might not understand they are the same thing and wonder if `maxSendAmount` is derived or transformed from `sendAmount` instead of set directly. If they are the same thing they should probably have the same name.

sabineschaller: It's all `debitAmount` now, even in the grant

.changeset/tender-ravens-breathe.md
sabineschaller: It's technically still a major change, but it feels so minor...

BlairCurrey: I think it should still be major since it's breaking, and I think we probably need to bump the `resource-server.yaml` version as well. Thats how we handled it when I renamed a field here: https://github.com/interledger/open-payments/pull/270. Although the spec version bump was a minor bump 🤔... don't quite remember that reasoning.

On the topic of it feeling minor, Max pointed to some relevant blog posts and hackernews discussion in this comment https://github.com/interledger/open-payments/pull/270#discussion_r1250797466. I don't see any quick, easy wins there and think just doing a major bump still makes the most sense but it is interesting reading.

adrianhopebailie (approved):
LGTM!

openapi/resource-server.yaml
```diff
@@ -576,7 +563,7 @@ paths:
...
                     paymentPointer: 'https://openpayments.guide/alice/'
                     receiver: 'https://openpayments.guide/aplusvideo/incoming-payments/45d495ad-b763-4882-88d7-aa14d261686e'
-                    sendAmount:
+                    debitAmount:
```
mkurapov, line 566: reference to sendAmount in the examples still

openapi/resource-server.yaml
```diff
@@ -641,7 +628,6 @@ paths:
                     - receiver
                     - sendAmount
                   additionalProperties: false
-
```
mkurapov, line 644: I think the quote still has `sendAmount`, are we keeping that?

mkurapov (commented):
Few references to `sendAmount` left

openapi/resource-server.yaml
sabineschaller: No, it's all `debitAmount` now, since the quote also has fees included.

mkurapov (commented):
Looks good :+1: Final comment, can you please update this final reference to `sendAmount`: https://github.com/interledger/open-payments/blob/main/packages/open-payments/README.md?plain=1#L182

PR #293, opened 2023-09-21
feat(open-payments): unauthenticated incoming payment get

BlairCurrey:
- adds public incoming payment return type to `/incoming-payments/{id}`
- removes sig params from `/incoming-payments/{id}`

## Context

fixes: https://github.com/interledger/rafiki/issues/1909

These changed were tested against rafiki with changes to return a public incoming-payment as well.


openapi/resource-server.yaml
```diff
@@ -684,9 +686,6 @@ paths:
         '404':
           description: Incoming Payment Not Found
       description: A client can fetch the latest state of an incoming payment to determine the amount received into the payment pointer.
-      parameters:
-        - $ref: '#/components/parameters/signature-input'
-        - $ref: '#/components/parameters/signature'
```
BlairCurrey, line 689: I tried to make these optional since they are still required to get the full incoming-payment details but it looks like this level of parameter (under `/get`) is required, which is why I removed them: https://stackoverflow.com/questions/35011192/how-to-define-an-optional-parameter-in-path-using-swagger

.changeset/silly-dragons-invite.md
```diff
@@ -0,0 +1,5 @@
+---
+'@interledger/open-payments': minor
+---
+
+Adds return public-incoming-payment returnt type to incoming payment get request which should be returned for unauthenticated requests to this endpoint
```
njlie, line 5: nit: typo on `returnt` -> `return`, if we don't change it here we'll have to change it on the bot PR.

BlairCurrey: fixed

openapi/resource-server.yaml
njlie: You could remove the `required: true` property from each of the signature components in the [shared components part of the spec](https://github.com/interledger/open-payments/blob/02ef4679da695fa47c32120309e73b4312c3dcab/openapi/resource-server.yaml#L1365-L1380), but that would make it optional for all resource paths...

Maybe there could be a one-off definition of the Signature parameters for this url that are optional. Or we could add `optional-signature` and `optional-signature-input` in the components section that are duplicates of `signature` and `signature-input` that optional requirements.

BlairCurrey: I added the `optional-signature` paramaters

njlie (approved):

openapi/resource-server.yaml
```diff
@@ -1365,5 +1381,21 @@ components:
         example: 'Signature-Input: sig1=("@method" "@target-uri" "content-digest" "content-length" "content-type");created=1618884473;keyid="gnap-rsa"'
       description: 'The Signature-Input field is a Dictionary structured field containing the metadata for one or more message signatures generated from components within the HTTP message.  Each member describes a single message signature.  The member''s key is the label that uniquely identifies the message signature within the context of the HTTP message.  The member''s value is the serialization of the covered components Inner List plus all signature metadata parameters identified by the label.  The following components MUST be included: - "@method" - "@target-uri" - "authorization".  When the message contains a request body, the covered components MUST also include the following: - "content-digest"  The keyid parameter of the signature MUST be set to the kid value of the JWK.      See [ietf-httpbis-message-signatures](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-message-signatures#section-4.1) for more details.'
       required: true
+    optional-signature:
```
njlie, line 1384: This is a nitpick that I don't feel the need to block over, but I wonder if the optional signature parameters could be reduced to a reference to the original required version but with an overridden `required` tag so there's less repeated text.

Something else to note is that Swagger defaults `required` to `false`, so perhaps the optional version should be the 'base' parameter and the required ones should extend those.

BlairCurrey: Good suggestion. I tried doing it like this:

```yaml
    signature-input:
      $ref: '#/components/parameters/optional-signature-input'
      required: true
```

Which seemed to work when running the code, except when generating the types it lost the non-schema details of the reference (so there was no comment with the description, and the key was `undefined` instead of `Signature-Input`, so I had to settle for just re-using the schema and duplicating the rest.

njlie (approved):

PR #305, opened 2023-09-26
feat: add wallet address to subresource requests

### LINKED ISSUE #304: Add wallet address as query/body parameter to OP resource GET/POST requests
In #290 wallet addresses were removed from resource endpoints' path parameters. They're yet to be replaced by some other property in the request to indicate what the address is supposed to be.

* For GET requests, include `walletAddress` in the query string.
* For POST requests, include `walletAddress` in the request body.

njlie:

* Adds the walletAddress field to request bodies for:
  * `POST /outgoing-payments`
  * `POST /incoming-payments`
  * `POST /incoming-payments/{id}/complete`
* Adds the wallet-address query parameter to requests for:
  * `GET /incoming-payments/{id}`
  * `GET /outgoing-payments/{id}`
  * `GET /incoming-payments`
  * `GET /outgoing-payments`

## Context

Fixes #304.

### DIFF
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -279,6 +279,7 @@ paths:
           $ref: '#/components/responses/403'
       description: List all incoming payments on the wallet address
       parameters:
+        - $ref: '#/components/parameters/wallet-address'
         - $ref: '#/components/parameters/cursor'
         - $ref: '#/components/parameters/first'
         - $ref: '#/components/parameters/last'
@@ -489,6 +490,7 @@ paths:
           $ref: '#/components/responses/403'
       description: List all outgoing payments on the wallet address
       parameters:
+        - $ref: '#/components/parameters/wallet-address'
         - $ref: '#/components/parameters/cursor'
         - $ref: '#/components/parameters/first'
         - $ref: '#/components/parameters/last'
@@ -657,7 +659,7 @@ paths:
       description: A client can fetch the latest state of an incoming payment to determine the amount received into the wallet address.
     parameters:
       - $ref: '#/components/parameters/id'
-      - $ref: '#/components/parameters/walletAddress'
+      - $ref: '#/components/parameters/wallet-address'
   '/incoming-payments/{id}/complete':
     post:
       summary: Complete an Incoming Payment
@@ -770,7 +772,7 @@ paths:
         - $ref: '#/components/parameters/signature'
     parameters:
       - $ref: '#/components/parameters/id'
-      - $ref: '#/components/parameters/walletAddress'
+      - $ref: '#/components/parameters/wallet-address'
   '/quotes/{id}':
     get:
       summary: Get a Quote
@@ -1305,7 +1307,7 @@ components:
         type: string
       description: Sub-resource identifier
       required: true
-    walletAddress:
+    wallet-address:
       name: wallet-address
       in: query
       schema:
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -154,6 +154,7 @@ paths:
             examples:
               Create incoming payment for $25 to pay invoice INV2022-02-0137:
                 value:
+                  walletAddress: 'https://openpayments.guide/alice/'
                   incomingAmount:
                     value: '2500'
                     assetCode: USD
@@ -697,6 +698,20 @@ paths:
           $ref: '#/components/responses/403'
         '404':
           description: Incoming Payment Not Found
+      requestBody:
+        content:
+          application/json:
+            schema:
+              type: object
+              additionalProperties: false
+              properties:
+                walletAddress:
+                  type: string
+                  description: The URL of the wallet address that should receive the incoming payment.
+            examples:
+              Complete incoming payment for https://openpayments.guide/alice:
+                value:
+                  walletAddress: 'https://openpayments.guide/alice/'
       description: |-
         A client with the appropriate permissions MAY mark a non-expired **incoming payment** as `completed` indicating that the client is not going to make any further payments toward this **incoming payment**, even though the full `incomingAmount` may not have been received.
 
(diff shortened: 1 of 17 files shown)


njlie:
Still need to add a changeset before this will be marked ready for review.

mkurapov (commented):
What do you think about always using the `walletAddress` in the query params, both in the GETs and the POSTs? To keep it consistent?

In any case, we should make wallet address always required + update the client package

njlie:
> What do you think about always using the `walletAddress` in the query params, both in the GETs and the POSTs? To keep it consistent?
> 
> In any case, we should make wallet address always required + update the client package

Do we need to support RESTful clients? I think those don't allow for query params in POSTs.

openapi/resource-server.yaml
```diff
@@ -690,6 +700,20 @@ paths:
...
+              type: object
+              additionalProperties: false
+              properties:
+                walletAddress:
+                  type: string
+                  description: The URL of the wallet address that should receive the incoming payment.
+            examples:
+              Complete incoming payment for https://openpayments.guide/alice:
+                value:
+                  walletAddress: 'https://openpayments.guide/alice/'
```
mkurapov, line 716: maybe this can just be the `ilp.money` wallet address one will be good so it works with Blair's PR regardless who merges first

openapi/resource-server.yaml
```diff
@@ -334,6 +339,9 @@ paths:
             schema:
               type: object
               properties:
+                walletAddress:
```
mkurapov, line 342: POST /quote request body should also contain `walletAddress`

mkurapov (commented):
> Do we need to support RESTful clients? I think those don't allow for query params in POSTs.
I wouldn't say that having query params in POST is necessarily _not_ RESTful, but I wasn't sure of the tradeoff between the weirdness of that vs needing to pass in the `walletAddress` differently depending on the method. We can just go with the more standard approach of no query params, was just curious to hear thoughts

Also, for the usage of the client, we can actually abstract that away by having it take in `walletAddress` as it's own parameter and do what we need to do internally in the client depending on the method:

```ts
client.incomingPayment.get({ url: '<some URL for the RS>', walletAddress: '<walletAddress>'...})  // uses wallet address in query params
client.incomingPayment.create({ url: '<some URL for the RS>', walletAddress: '<walletAddress', accessToken: '' }, { expiresAt: ... }) // uses wallet address in request body
```

njlie:
> > Do we need to support RESTful clients? I think those don't allow for query params in POSTs.
> 
> I wouldn't say that having query params in POST is necessarily _not_ RESTful, but I wasn't sure of the tradeoff between the weirdness of that vs needing to pass in the `walletAddress` differently depending on the method. We can just go with the more standard approach of no query params, was just curious to hear thoughts
> 
> Also, for the usage of the client, we can actually abstract that away by having it take in `walletAddress` as it's own parameter and do what we need to do internally in the client depending on the method:
> 
> ```ts
> client.incomingPayment.get({ url: '<some URL for the RS>', walletAddress: '<walletAddress>'...})  // uses wallet address in query params
> client.incomingPayment.create({ url: '<some URL for the RS>', walletAddress: '<walletAddress', accessToken: '' }, { expiresAt: ... }) // uses wallet address in request body
> ```

Yeah, I'm thinking the no query parameters in POST will work. If someone asks "why did they do it this way?", then it may be a stronger argument to point to that dogma than a personal preference for consistency

packages/open-payments/src/client/index.ts
```diff
@@ -58,11 +58,14 @@ interface AuthenticatedRequestArgs {
...
 }
-export interface ResourceRequestArgs
-  extends UnauthenticatedResourceRequestArgs,
+
+export interface GrantOrTokenRequestArgs
+  extends UnauthenticatedRequestArgs,
     AuthenticatedRequestArgs {}
 
-export interface CollectionRequestArgs extends AuthenticatedRequestArgs {
+export interface ResourceOrCollectionRequestArgs
```
mkurapov, line 66: Just a thought I had, should we change the name of the `url` argument to be instead `resourceServerUrl` for the POST requests (since we end up appending the resource name, e.g. /incoming-payment), to be more clear that we _will_ be appending the resource name?

mkurapov (approved):
LGTM, not sure if this should be merged before or after https://github.com/interledger/open-payments/pull/307, however

mkurapov:
Do we need to return the `resourceServerUrl` in the wallet address lookup?

njlie:
> Do we need to return the `resourceServerUrl` in the wallet address lookup?

I think https://github.com/interledger/open-payments/pull/305#discussion_r1352855227 also covers this. I think this version of Open Payments doesn't need to account for resource servers differing from the wallet address until Adrian's PR. Just need to make sure it's captured in an issue or in the scope of that draft PR

packages/open-payments/src/client/index.ts
sabineschaller: Sorry, this 👆was supposed to be a comment to you question below.

openapi/resource-server.yaml
```diff
@@ -650,6 +671,7 @@ paths:
       description: A client can fetch the latest state of an incoming payment to determine the amount received into the wallet address.
     parameters:
       - $ref: '#/components/parameters/id'
+      - $ref: '#/components/parameters/wallet-address'
```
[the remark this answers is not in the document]
njlie: @sabineschaller Is it the case also for completing an incoming payment (https://github.com/interledger/open-payments/pull/305/files#diff-9debc2038853698222051b2c2a428cee93e9dda6af6076de559971933879580dR722-R725) since the path still contains that unique id?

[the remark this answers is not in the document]
sabineschaller: Yes, indeed.

sabineschaller (approved):
LGTM

PR #307, opened 2023-09-27
feat(open-payment): add unauthenticated payment get

BlairCurrey:
- adds `incomingPayment.get` method to unauthenticated client

## Context
- fixes: https://github.com/interledger/open-payments/issues/306


packages/open-payments/src/client/incoming-payment.ts
```diff
@@ -89,9 +90,34 @@ export const createIncomingPaymentRoutes = (
   }
 }
 
+export interface UnauthenticatedIncomingPaymentRoutes {
+  get(args: UnauthenticatedResourceRequestArgs): Promise<IncomingPayment>
```
mkurapov, line 94: This should actually be a new type of IncomingPayment we return, something like `PublicIncomingPayment` that only returns what's defined in the spec

njlie (approved):
LGTM

packages/open-payments/README.md
```diff
@@ -34,7 +34,7 @@ This package exports two clients, an `UnauthenticatedClient` and an `Authenticat
 ### `UnauthenticatedClient`
 
 This client allows making requests to access publicly available resources, without needing authentication.
-The three available resources are [Wallet Addresses](https://docs.openpayments.guide/reference/get-wallet-address), [Wallet Address Keys](https://docs.openpayments.guide/reference/get-wallet-address-keys), and [ILP Stream Connections](https://docs.openpayments.guide/reference/get-ilp-stream-connection).
```
mkurapov, line 37: We should still keep the keys in here

packages/open-payments/src/client/index.ts
```diff
@@ -124,6 +127,11 @@ export const createUnauthenticatedClient = async (
       axiosInstance,
       openApi: resourceServerOpenApi,
       logger
+    }),
+    incomingPayment: createUnauthenticatedIncomingPaymentRoutes({
```
mkurapov, line 131: it would be great if the public incoming payment was also possible to be fetched using the authenticated client (either through overloading `incomingPayment.get` somehow, or just adding `incomingPayment.getPublic()` (or something else maybe)?

BlairCurrey: Something like this seems reasonable. I believe I can just remove the authorization header (via not passing in an `accessToken`) to get the public version.
```ts
export const createIncomingPaymentRoutes = (
  deps: RouteDeps
): IncomingPaymentRoutes => {

    ...
    return {
        ...
        
        getPublic: (args: ResourceRequestArgs) => {
          const { accessToken, ...argsWithoutAccessToken } = args
          return getPublicIncomingPayment(
            { axiosInstance, logger },
            argsWithoutAccessToken,
            getPublicIncomingPaymentOpenApiValidator
          )
        }
        
        ...
    }
    ...
    
```

BlairCurrey: also considered the function overloading with an implementation like:

```ts
export const getIncomingPaymentWIP = async (
  deps: BaseDeps,
  args: ResourceRequestArgs | UnauthenticatedResourceRequestArgs,
  validateOpenApiResponse:
    | ResponseValidator<IncomingPaymentWithPaymentMethods>
    | ResponseValidator<PublicIncomingPayment>
) => {
  if ('accessToken' in args) {
    return getIncomingPayment(deps, args, validateOpenApiResponse as ResponseValidator<IncomingPaymentWithPaymentMethods>)
  } else {
    getPublicIncomingPayment(deps, args, validateOpenApiResponse as ResponseValidator<PublicIncomingPayment>)
  }
}
```

But then I think we're back to a union return type (an incoming payment OR public incoming payment).

packages/open-payments/README.md
BlairCurrey: fixed

mkurapov (approved):

Issue #303, opened 2023-09-25
Add `baseUrl` to OP spec


njlie:
In #290 we removed the wallet address from resource URLs, but that left the spec without any sort of base URL, since the wallet address _was_ the base url for most endpoints. A base URL of some kind should be reinstated. 

`rafiki.money` might be a good example value.

PR #308, opened 2023-09-27
fix(openapi): add base url

BlairCurrey:
- adds `rafiki.money` as base path

image

## Context
- fixes https://github.com/interledger/open-payments/issues/303

The issues suggests maybe using rafiki.money as the base url but I opted for openpayments.guide because that's what it originally was. https://github.com/interledger/open-payments/blob/5d39b2ef5f202e6378555b790730f9465ecd6f97/openapi/resource-server.yaml#L35C3-L39 This makes the base url consistent with existing examples.

### DIFF
--- a/.changeset/loud-snakes-melt.md
+++ b/.changeset/loud-snakes-melt.md
@@ -1,5 +1,0 @@
----
-'@interledger/open-payments': major
----
-
-Changes base path from / to /openpayments.guide
--- a/.changeset/loud-snakes-melt.md
+++ b/.changeset/loud-snakes-melt.md
@@ -1,0 +1,5 @@
+---
+'@interledger/open-payments': major
+---
+
+Changes base path from / to /openpayments.guide
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -32,8 +32,8 @@ info:
   contact:
     email: tech@interledger.org
 servers:
-  - url: 'https://openpayments/'
-    description: 'Server for wallet address subresources or Connection resources (ie https://openpayments.guide/)'
+  - url: 'https://openpayments.guide'
+    description: 'Server for wallet address subresources or Connection resources (ie https://openpayments.guide/alice)'
 tags:
   - name: wallet-address
     description: wallet address operations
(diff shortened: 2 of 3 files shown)


openapi/resource-server.yaml
```diff
@@ -112,7 +112,7 @@ paths:
                 New Incoming Payment for $25:
                   value:
                     id: 'https://openpayments.guide/incoming-payments/08394f02-7b7b-45e2-b645-51d04e7c330c'
-                    walletAddress: 'https://openpayments.guide/alice/'
```
njlie, line 115: I believe `walletAddress` values should still have `/alice`. That seems like the case to me but I could be wrong @mkurapov / @sabineschaller

.changeset/loud-snakes-melt.md
```diff
@@ -0,0 +1,5 @@
+---
+'@interledger/open-payments': major
```
njlie, line 2: I don't think this constitutes a major change. In fact, I'm not sure this changes anything in the `open-payments` package itself at all.

openapi/resource-server.yaml
```diff
@@ -32,8 +32,8 @@ info:
   contact:
     email: tech@interledger.org
 servers:
-  - url: '/'
-    description: 'Server for wallet address subresources or Connection resources (ie https://openpayments.guide/alice)'
+  - url: 'https://openpayments/'
```
mkurapov, line 35: Not sure if I have a preference, but `rafiki.money` might be nice to have if we are going to include some code snippets in the documentation that _do_ use `rafiki.money`. I would connect with Radu /Melissa/Mohammed for this

openapi/resource-server.yaml
mkurapov: Yeah, the wallet address remains the same, just the baseUrl changes

.changeset/loud-snakes-melt.md
BlairCurrey: removed

openapi/resource-server.yaml
BlairCurrey: I will change to rafiki.money. Melissa confirmed that there will be code snippets in the docs that are intended to be used with rafiki.money. https://github.com/interledger/open-payments-snippets She also noted the openpayments.guide domain might change as well but in any case it seems like the rafiki.money url will be more useful.

openapi/resource-server.yaml
BlairCurrey: this and similar are now `https://rafiki.money/alice/`

mkurapov (approved):

njlie (approved):

PR #320, opened 2023-10-10
fix(openapi): allow additional properties in wallet address response

DonChangfoot:

Allow extra fields to be returned in the response to looking up a wallet address.

## Context
Fynbos returns a superset of the specified wallet address response. This is currently causing the lookup to fail.

### DIFF
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -54,6 +54,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/wallet-address'
+                additionalProperties: true
               examples:
                 Get wallet address for $ilp.rafiki.money/alice:
                   value:


openapi/resource-server.yaml
```diff
@@ -54,6 +54,7 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/wallet-address'
+                additionalProperties: true
```
mkurapov, line 57: Does this override the `additionalProperties: false` on the $ref: https://github.com/interledger/open-payments/blob/00d479191ca7b8e277b36dc6bec480f5d31a19ef/openapi/resource-server.yaml#L796-L800

mkurapov: yeah, `additionalProperties: true` on the wallet-address itself is ok IMO

sabineschaller:
Duplicate of #326

PR #326, opened 2023-10-13
allow additional properties in wallet address response

matdehaast:


sabineschaller (approved):
We should close Don's PR with this one.

matdehaast:
> LGTM, just update the type (see PR checks).
> 
> Also, we should close Don's PR with this one.

He was away and I wanted to try get this in quicker. Will update the types now

Issue #330, opened 2023-10-13
Remove addtionalProperties: false from incoming-payment type


BlairCurrey:
This matches how we previously handled `incoming-payment` and `incoming-payment-with-connection` (now `incoming-payment-with-methods`) and prevents our jest openapi spec checks from incorrectly failing. 

Openapi (or at least some openapi tooling) does not handle the use of `allOf` and `addtionalProperties: false` as we might expect.

PR #331, opened 2023-10-13
fix(openapi): remove additional properties

BlairCurrey:
## Context
Matches what we had before the connection -> payment methods changes that works with our jest openapi checks.
- fixes: #330

### DIFF
--- a/.changeset/lovely-feet-dance.md
+++ b/.changeset/lovely-feet-dance.md
@@ -1,0 +1,5 @@
+---
+'@interledger/open-payments': patch
+---
+
+removed additionalProperties: false from incoming-payment and added unevaluatedProperties: false
--- a/.changeset/lovely-feet-dance.md
+++ b/.changeset/lovely-feet-dance.md
@@ -2,4 +2,4 @@
 '@interledger/open-payments': patch
 ---
 
-removed additionalProperties: false from incoming-payment and added unevaluatedProperties: false
+moves additionalProperties: false from incoming-payment type to GET /incoming-payments response
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -883,6 +883,7 @@ components:
           expiresAt: '2022-04-12T23:20:50.52Z'
           createdAt: '2022-03-12T23:20:50.52Z'
           updatedAt: '2022-03-12T23:20:50.52Z'
+      unevaluatedProperties: false
       properties:
         id:
           type: string
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -194,6 +194,7 @@ paths:
                     type: array
                     items:
                       $ref: '#/components/schemas/incoming-payment'
+                      additionalProperties: false
               examples:
                 forward pagination:
                   value:
@@ -883,7 +884,6 @@ components:
           expiresAt: '2022-04-12T23:20:50.52Z'
           createdAt: '2022-03-12T23:20:50.52Z'
           updatedAt: '2022-03-12T23:20:50.52Z'
-      unevaluatedProperties: false
       properties:
         id:
           type: string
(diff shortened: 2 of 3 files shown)


openapi/resource-server.yaml
```diff
@@ -883,7 +883,6 @@ components:
           expiresAt: '2022-04-12T23:20:50.52Z'
           createdAt: '2022-03-12T23:20:50.52Z'
           updatedAt: '2022-03-12T23:20:50.52Z'
-      additionalProperties: false
```
mkurapov, line 886: would `unevaluatedProperties` work here instead?

packages/openapi/src/middleware.test.ts
```diff
@@ -186,7 +186,6 @@ describe('OpenAPI Validator', (): void => {
     test.each`
       status | body                                                                    | message                                                           | description
       ${202} | ${{}}                                                                   | ${'An unknown status code was used and no default was provided.'} | ${'status code'}
-      ${201} | ${{ ...body, invalid: 'field' }}                                        | ${'response must NOT have additional properties: invalid'}        | ${'body with additional property'}
```
BlairCurrey, line 189: I expected this since it's no longer restricting additional properties, but I'm surprised this case in a previous test is still working with `unevaluatedProperties: false` (and that this case wouldn't behave the same way)

https://github.com/interledger/open-payments/blob/32e089c431cbb17f93e1a9d5c218e84d0618a593/packages/openapi/src/middleware.test.ts#L133

mkurapov: The 
```
${{ additionalProp: 'disallowed' }} | ${'body must NOT have additional properties: additionalProp'} | ${'invalid additional property'}
```

 test looks like its for the incoming-payments _request_ instead?

openapi/resource-server.yaml
```diff
@@ -883,7 +883,7 @@ components:
...
+      unevaluatedProperties: false
```
mkurapov, line 886: is there a way we can test this one?

packages/openapi/src/middleware.test.ts
BlairCurrey: Ah, yes. that's it.

openapi/resource-server.yaml
BlairCurrey: After some more testing, this is actually not doing anything when adding an additional property to the response. It does not cause any error. From reading [the underlying json-schema docs](https://json-schema.org/draft/2020-12/json-schema-core#name-unevaluatedproperties) and [other usages](https://github.com/search?q=path%3A*.yaml+OR+path%3A*.yml+%22unevaluatedProperties%22&type=code&p=2) it seems like this actually should go in the `incoming-payment-with-methods` (where the `allOf` is), but that didn't work either.

packages/openapi/src/middleware.test.ts
```diff
@@ -172,6 +172,41 @@ describe('OpenAPI Validator', (): void => {
...
+              updatedAt: '2022-04-01T10:24:36.11Z',
+              additionalProp: 'disallowed'
+            }
+          ]
+        }
+      })
+      await expect(validateListMiddleware(ctx, next)).rejects.toMatchObject({
+        status: 500,
+        message:
+          'response.result.0 must NOT have additional properties: additionalProp'
```
mkurapov, line 205: I'm guessing this is a test for an older spec?

mkurapov (approved):

.changeset/lovely-feet-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+'@interledger/open-payments': patch
+---
+
+removed additionalProperties: false from incoming-payment and added unevaluatedProperties: false
```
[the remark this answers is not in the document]
BlairCurrey: updated

packages/openapi/src/middleware.test.ts
BlairCurrey: When I pushed this test up it was passing locally but failed here until I update my branch (the gh action used the new one from main somehow?). I added this because I moved `additionalProperties: false` to the `GET /incoming-payments` response (the list) from the `incoming-payment` types itself. So it's testing that, which should preserve existing functionality.

mkurapov (approved):

PR #341, opened 2023-10-19
fix: resource server spec

sabineschaller:

This fixes validation error and warnings.

## Context

When @huijing parsed the spec, she picked up a validation error that our GH workflow did not pick up.


huijing (approved):
All good from the Swagger Editor validator front :)

PR #354, opened 2023-10-26
feat: open payments client can now take in a file path as an argument to `privateKey`

### LINKED ISSUE #329: Allow passing in a file path for setting the privateKey in the Open Payments client
Instead of only allowing `privateKey: KeyLike` as an argument when calling `createAuthenticatedClient`, we should allow a file name to be passed in as an alternative as well. If using the file name, we should use the `http-signature-utils` `parseOrProvisionKey` method as a way to load the key into the client.

mkurapov:

Instead of taking in a `privateKey: KeyLike` when creating an open payments client using `createAuthenticatedClient`, use: `privateKeyFilePath: string` instead, so the client does the file key parsing during initialization. 

This allows users of the client to not have to separately import and call `parseKey` from `http-signature-utils`. 

## Context

Fixes #329


packages/http-signature-utils/src/utils/key.ts
```diff
@@ -1,44 +1,66 @@
...
+export function parseOrProvisionKey(keyFile?: string): crypto.KeyObject {
+  if (keyFile) {
+    try {
+      return parseKey(keyFile)
+    } catch (error) {
+      console.warn(
+        `parseOrProvisionKey: could not parse key (${
+          error instanceof Error ? error.message : 'unknown error'
+        }). Generating new key.`
+      )
```
mkurapov, line 47: This is "basically" the old behaviour, except it will re-generate a key in any error, not just if the key is a different curve.

I'm not sure if I like this behaviour in general, since this fails silently if the key that you expected to load didn't. I'm not sure about the console.logs either in a published package.

Would be great to get some thoughts on this

packages/open-payments/src/client/index.ts
```diff
@@ -86,8 +87,28 @@ export interface CollectionRequestArgs
 const createDeps = async (
   args: Partial<CreateAuthenticatedClientArgs>
 ): Promise<ClientDeps> => {
+  const logger = args?.logger ?? createLogger({ name: 'Open Payments Client' })
```
mkurapov, line 90: adding `name` so its more clear in the logs where the error is from

packages/http-signature-utils/README.md
```diff
@@ -19,6 +19,18 @@ npm install @interledger/http-signature-utils
...
+```ts
+const key = parseKey('/PATH/TO/private-key.pem')
+```
+
+Generate a private Ed25519 key:
+
+```ts
+const key = provisionKey('/PATH_TO_SAVE_KEY_IN')
+```
+
```
mkurapov, line 33: I followed the previous naming, but I wouldn't mind using `loadKey` and `generateKey` as well

packages/open-payments/src/client/index.ts
```diff
@@ -86,8 +87,28 @@ export interface CollectionRequestArgs
...
+
+  let privateKey: KeyLike | undefined
+
+  try {
+    if (args.privateKeyFilePath) {
+      privateKey = parseKey(
+        path.resolve(process.cwd(), args.privateKeyFilePath)
```
mkurapov, line 97: Allows taking in relative path or absolute.

packages/http-signature-utils/src/utils/key.ts
BlairCurrey: Maybe we should expect the caller to choose between these options:
- parse key (and fail if it cant)
- generate a key (perhaps optional)
- try to parse a key, and regenerate one if it fails

Which we already have as far as I can tell (`parseKey`, `provisionKey`, `parseOrProvisionKey`). I think the function signatures communicate this pretty well already. Maybe we could make the docs for `parseOrProvisionKey` more explicit ("will generate a new key if the parse fails") and include it in a docstring but it's behaving how I would expect personally.

If we really need to communicate parse failures in `parseOrProvisionKey` maybe we return an object like `{key: crypto.KeyObject, parseSucceeded: boolean}`. I guess I'm not really about the console.logs but I feel like we shouldn't be doing it like this. If we wanted to log I imagine we should support passing in popular loggers (winston, bunyon, pino, etc.) and using those (but totally overkill here).

BlairCurrey: Also I dont have the full context on the previous behavior, but distinguishing between a non EdDSA-Ed25519 private key and something unparsable seems like an unimportant distinction to me.

packages/http-signature-utils/README.md
BlairCurrey: I don't think you necessarily need to change it but `loadKey` and `generateKey` are clearer to me.

raducristianpopa:
Is there a possibility to have backward compatibility for this?

packages/http-signature-utils/src/utils/key.ts
mkurapov: Sounds good Blair, agreed. I will add some doc strings as well

packages/http-signature-utils/README.md
mkurapov: Yep, with you. Made this a major lib bump and updated the naming

.changeset/kind-singers-roll.md
```diff
@@ -0,0 +1,6 @@
+---
+'@interledger/open-payments': major
+---
+
+- `createAuthenticatedClient` can now also load a key using a path to the private key file as an argument to `privateKey`
+- `walletAddress` is required in the incoming payment creation request
```
mkurapov, line 6: Unfourtunately had to make this a major bump because of the `walletAddress` being required now

packages/open-payments/src/client/index.ts
```diff
@@ -145,8 +192,8 @@ export const createUnauthenticatedClient = async (
 
 export interface CreateAuthenticatedClientArgs
   extends CreateUnauthenticatedClientArgs {
-  /** The private EdDSA-Ed25519 key with which requests will be signed */
-  privateKey: KeyLike
+  /** The path to the file containing the EdDSA-Ed25519 key with which requests will be signed */
+  privateKey: string | KeyLike
```
mkurapov, line 196: a  bit redundant because KeyLike is just a type of `string | Buffer | KeyObject` but I like the explicitness

packages/http-signature-utils/src/utils/key.ts
```diff
@@ -1,44 +1,62 @@
...
+
+export function generateKey(dir: string): crypto.KeyObject {
   const keypair = crypto.generateKeyPairSync('ed25519')
-  if (!fs.existsSync(TMP_DIR)) {
-    fs.mkdirSync(TMP_DIR)
+  if (!fs.existsSync(dir)) {
+    fs.mkdirSync(dir)
   }
   fs.writeFileSync(
-    `${TMP_DIR}/private-key-${new Date().getTime()}.pem`,
```
mkurapov, line 27: @sabineschaller what was the use case for actually saving the key to a file when calling `parseOrProvisionKey` ?

mkurapov: In Rafiki, we call `parseOrProvisionKey` during backend app config, and use the generated KeyObject directly if the file for the key wasn't provided. We don't really end up using the file of the key

packages/openapi/src/middleware.test.ts
```diff
@@ -244,7 +244,7 @@ describe('OpenAPI Validator', (): void => {
           {}
         )
         addTestSignatureHeaders(ctx)
-        ctx.request.body = {}
+        ctx.request.body = { walletAddress: WALLET_ADDRESS }
```
mkurapov, line 247: was failing tests

BlairCurrey (approved):

raducristianpopa (approved):

BlairCurrey (approved):

PR #369, opened 2023-11-07
mi-quote-create-debit-receive-amount

hajjimo:


melissahenderson (changes requested):
The "Create a fixed-send amount quote" has a fourth chunk in the source file that's not showing in the doc. 

Also take a look at the second paragraph in the overview. I believe it needs to be revised since it was originally written when we only had the "fixed send" script.

melissahenderson (approved):

PR #371, opened 2023-11-08
Mi recurring payments guide

hajjimo:


docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -0,0 +1,274 @@
...
+- TypeScript
+- NPM
+
+### Dependencies
+
+- <LinkOut href='https://github.com/interledger/open-payments/tree/main/packages/open-payments'>
+    Open Payments SDK
+  </LinkOut>
+  <p></p>
+  ### Endpoints
```
melissahenderson, line 23: There's a space before "###" so the heading is indented.

PR #377, opened 2023-11-20
docs: Resolves #376

### LINKED ISSUE #376: Improve 'before you begin' section in code snippets user docs
Each code snippets page has a Before You Begin section that contains multiple expands for signing up with/setting up Rafiki.money. However, after going through the steps once, you don't need to do them again. Instead of having the same set of expands on every page, let's create a stand-alone page (or pages) and link to the content from the code snippets pages.

- Decide whether to create one 'before you begin' page for all content or one page per step within the Code Snippets section
- Update each code snippet page by removing the associated 'import' and 'disclosure' markdown
- To ensure we use the same lang string and link(s) on each code snippet page, create the new 'before you begin' content within a partial and add to each page

melissahenderson:

See title

## Context

Closes #376


melissahenderson:
Good 'ol Playwright...

huijing (approved):
Hope nothing explodes.

PR #379, opened 2023-11-22
fix: add the `useHttp` option to the missing places

### LINKED ISSUE #378: [BUG] `useHttp` is not passed to request functions and routes factories
## Bug Report

**Describe the bug**

`useHttp` is not passed to the routes factory functions and the request function (`get`, `post`, `delete`).

**To Reproduce**

Steps and commands to reproduce the behavior. Please provide clear code snippets that always reproduce the bug.
Running this script, with a patched `@interledger/open-payments` package to see if the `useHttp` option is being used or not:
```ts
import { createUnauthenticatedClient } from "@interledger/open-payments";

const client = await createUnauthenticatedClient({
    useHttp: true,
});

await client.walletAddress.get({
    url: "https://localhost:3010/client",
});
```

will result in:
```
[PATCHED] 'useHttp' in 'checkUrlProtocol': undefined
{"level":50,"time":1700569881397,"pid":32795,"hostname":"Radus-MacBook-Pro-14.local","name":"Open Payments Client","errorDescription":"write EPROTO C01EBDDB01000000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:../deps/openssl/openssl/ssl/record/ssl3_record.c:355:\n","msg":"Error making Open Payments GET request"}
/Users/radu/dev/web-monetization-script/node_modules/.pnpm/@interledger+open-payments@6.1.0_patch_hash=rqtnfdefv5gtohzzymff6lzqby/node_modules/@interledger/open-payments/dist/client/requests.js:119
```

**Expected behavior**

If `useHttp` is set to `true`, all URLs should be converted to `http`.

raducristianpopa:
## Context

Closes #378.

### DIFF
--- a/packages/open-payments/src/client/index.ts
+++ b/packages/open-payments/src/client/index.ts
@@ -30,7 +30,7 @@ export * from './error'
 export interface BaseDeps {
   axiosInstance: AxiosInstance
   logger: Logger
-  useHttp?: boolean
+  useHttp: boolean
 }
 
 interface ClientDeps extends BaseDeps {
@@ -118,9 +118,10 @@ const parseKey = (
   }
 }
 
-const createDeps = async (
-  args: Partial<CreateAuthenticatedClientArgs>
-): Promise<ClientDeps> => {
+const createDeps = async ({
+  useHttp = false,
+  ...args
+}: Partial<CreateAuthenticatedClientArgs> = {}): Promise<ClientDeps> => {
   const logger = args?.logger ?? createLogger({ name: 'Open Payments Client' })
   if (args.logLevel) {
     logger.level = args.logLevel
@@ -158,7 +159,8 @@ const createDeps = async (
     axiosInstance,
     resourceServerOpenApi,
     authServerOpenApi,
-    logger
+    logger,
+    useHttp: useHttp
   }
 }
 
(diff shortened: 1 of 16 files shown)


packages/open-payments/src/client/index.ts
```diff
@@ -118,9 +121,36 @@ const parseKey = (
   }
 }
 
-const createDeps = async (
-  args: Partial<CreateAuthenticatedClientArgs>
-): Promise<ClientDeps> => {
+const createUnauthenticatedDeps = async ({
```
raducristianpopa, line 124: Thought it would be okay to separate the unauthenticated dependencies from the authenticated ones. This way, when initializing an unauthenticated client, the authentication server OpenAPI schema will not be loaded.

packages/open-payments/src/client/index.ts
```diff
@@ -30,11 +30,14 @@ export * from './error'
 export interface BaseDeps {
   axiosInstance: AxiosInstance
   logger: Logger
-  useHttp?: boolean
+  useHttp: boolean
```
raducristianpopa, line 33: Made the `useHttp` dependency required so we can get an error whenever we are not passing it.

packages/open-payments/src/client/token.ts
```diff
@@ -30,18 +26,14 @@ export const rotateToken = async (
 }
 
 export const revokeToken = async (
-  deps: RouteDeps,
+  deps: BaseDeps,
```
raducristianpopa, line 29: Token functions were using `RouteDeps` instead of `BaseDeps` which as well.

mkurapov (approved):
Thanks!

Issue #412, opened 2024-01-12
Enable the option to provide a custom request interceptor for the authenticated client


raducristianpopa:
## Context

Given the lack of full browser compatibility for our algorithm (Ed25519), transitioning to the SubtleCrypto API is not feasible. The Axios instance is using a request interceptor to create HTTP signatures using `@interledger/http-signature-utils`. Because this process relies on Node’s crypto module for signing requests, the client cannot be directly utilized in the WM extension.

## Todos
- [ ] For the authenticated client, allow users to provide a custom request interceptor rather than the private key and key ID. This custom interceptor will then be registered and replace the built-in interceptor for generating signatures.

raducristianpopa:
For the WM extension, we initially needed a solution to enable running the SDK in a browser environment without having to recreate all its functionalities. However, this approach has now changed. We have opted to use the SDK on a server that functions as a proxy. This server is responsible for generating HTTP signatures and forwarding requests to an ASE's authentication or resource server. 

Closing this issue as not planned for now.

PR #417, opened 2024-02-06
feat(sig-utils): add browser support

### LINKED ISSUE #416: Drop the usage of the node crypto module for `http-signature-utils`
## Context
Currently, using the SDK outside of a Node environment is impractical, primarily due to the use on Node's crypto module for generating key pairs using the Ed25519 algorithm. The [SubtleCrypto API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto), while accessible in both browser and Node (v18 or earlier - with polyfill, v20 or later - without polyfill) environments, does not yet offer support for Ed25519 or is experimental:

- In Chromium-based browsers, the algorithm is accessible by activating the `Experimental Web Platform Features` flag ([implementation status](https://github.com/WICG/webcrypto-secure-curves/issues/20));
- Firefox does not support the algorithm;
- In Node, the Ed25519 algorithm is still considered [experimental](https://nodejs.org/dist/latest-v20.x/docs/api/webcrypto.html#ed25519ed448x25519x448-key-pairs);
  
## Todos
- [ ] Update Node version to 20 (with backwards compatibility)
- [ ] Use `@noble/ed25519` to generate the key pair;
- [ ] Take advantage of an ASN.1 decoder to validate the algorithm of a private key when loading it;
- [ ] Update `http-message-signatures` to it's latest version (using 0.1.2 - latest 1.0.4);
- [ ] Make sure that importing private keys that were generated with the Node crypto module or other libraries work as well;

raducristianpopa:
## Context

Closes #416.

### DIFF
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -24,7 +24,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v3
         with:
-          node-version: 18
+          node-version: 20
 
       - uses: pnpm/action-setup@v2
         name: Install pnpm
--- a/.github/workflows/env-setup/action.yaml
+++ b/.github/workflows/env-setup/action.yaml
@@ -9,7 +9,7 @@ runs:
         version: 7
     - uses: actions/setup-node@v3
       with:
-        node-version: '18'
+        node-version: '20'
         cache: 'pnpm'
     - name: Install dependencies
       shell: bash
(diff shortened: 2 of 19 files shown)


raducristianpopa (commented):
This draft pull request initiates the addition of browser support to the library. Future pull requests will target this branch to incrementally update all parts.

Given the current challenge - no browser support or experimental feature for our algorithm, would it fit best to release the libraries with browser compatibility under a specific tag? This would enable the WM extension to take advantage of the Open Payments SDK without replicating its capabilities and to prevent vulnerabilities from being introduced into the stable releases.

raducristianpopa:
Closing in favor of #412 and #420.

PR #421, opened 2024-02-09
feat(sig-utils): upgrade `http-message-signatures` to the most recent version

### LINKED ISSUE #420: Update `http-message-signatures` version
## Context
We aree currently using an outdated version of the `http-message-signatures` library.
## Todos
- [ ] Upgrade the `http-message-signatures` to the latest version and address any breaking changes.

raducristianpopa:
Update `http-message-signatures` to its latest version and fix breaking changes.

## Context

Closes #420.

### DIFF
--- a/packages/http-signature-utils/src/utils/signatures.ts
+++ b/packages/http-signature-utils/src/utils/signatures.ts
@@ -38,7 +38,11 @@ export const createSignatureHeaders = async ({
       params: ['alg', 'keyid', 'created'],
       fields: components
     },
-    request
+    {
+      method: request.method,
+      url: request.url,
+      headers: request.headers
+    }
   )
 
   return {
--- a/packages/http-signature-utils/src/utils/signatures.ts
+++ b/packages/http-signature-utils/src/utils/signatures.ts
@@ -35,7 +35,7 @@ export const createSignatureHeaders = async ({
     {
       key: signingKey,
       name: 'sig1',
-      params: ['alg', 'keyid', 'created'],
+      params: ['keyid', 'created'],
       fields: components
     },
     {
--- a/packages/http-signature-utils/src/utils/signatures.ts
+++ b/packages/http-signature-utils/src/utils/signatures.ts
@@ -1,10 +1,9 @@
-import { sign, KeyLike, BinaryLike } from 'crypto'
-import {
-  httpis as httpsig,
-  Algorithm,
-  RequestLike,
-  Signer
-} from 'http-message-signatures'
+import { type KeyLike } from 'crypto'
+import { httpbis, createSigner, Request } from 'http-message-signatures'
+
+export interface RequestLike extends Request {
+  body?: string
+}
 
 export interface SignOptions {
   request: RequestLike
@@ -15,13 +14,6 @@ export interface SignOptions {
 export interface SignatureHeaders {
   Signature: string
   'Signature-Input': string
-}
-
-const createSigner = (privateKey: KeyLike): Signer => {
-  const signer = async (data: BinaryLike) =>
-    sign(null, data as Buffer, privateKey)
-  signer.alg = 'ed25519' as Algorithm
-  return signer
 }
 
 export const createSignatureHeaders = async ({
@@ -36,15 +28,19 @@ export const createSignatureHeaders = async ({
   if (request.body) {
     components.push('content-digest', 'content-length', 'content-type')
   }
-  const { headers } = await httpsig.sign(request, {
-    components,
-    parameters: {
-      created: Math.floor(Date.now() / 1000)
+
+  const signingKey = createSigner(privateKey, 'ed25519', keyId)
+
+  const { headers } = await httpbis.signMessage(
+    {
+      key: signingKey,
+      name: 'sig1',
+      params: ['alg', 'keyid', 'created'],
+      fields: components
     },
-    keyId,
-    signer: createSigner(privateKey),
-    format: 'httpbis'
-  })
+    request
+  )
+
   return {
     Signature: headers['Signature'] as string,
     'Signature-Input': headers['Signature-Input'] as string
--- a/packages/http-signature-utils/src/utils/validation.ts
+++ b/packages/http-signature-utils/src/utils/validation.ts
@@ -1,5 +1,5 @@
 import * as crypto from 'crypto'
-import { RequestLike } from 'http-message-signatures'
+import { RequestLike } from './signatures'
 import { verifyContentDigest } from 'httpbis-digest-headers'
 import { importJWK } from 'jose'
 import { JWK } from './jwk'
--- a/.changeset/light-rabbits-rhyme.md
+++ b/.changeset/light-rabbits-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@interledger/http-signature-utils': patch
+---
+
+Upgrade the `http-message-signatures` package to the most recent version.
--- a/packages/http-signature-utils/package.json
+++ b/packages/http-signature-utils/package.json
@@ -24,7 +24,7 @@
     "prepack": "pnpm build"
   },
   "dependencies": {
-    "http-message-signatures": "^0.1.2",
+    "http-message-signatures": "^1.0.4",
     "httpbis-digest-headers": "^1.0.0",
     "jose": "^4.13.1",
     "uuid": "^9.0.0"
--- a/packages/http-signature-utils/src/index.ts
+++ b/packages/http-signature-utils/src/index.ts
@@ -6,7 +6,6 @@ export {
   loadOrGenerateKey,
   loadBase64Key
 } from './utils/key'
-export { createSignatureHeaders } from './utils/signatures'
+export { createSignatureHeaders, type RequestLike } from './utils/signatures'
 export { validateSignatureHeaders, validateSignature } from './utils/validation'
 export { generateTestKeys, TestKeys } from './test-utils/keys'
-export { RequestLike } from 'http-message-signatures'
--- a/packages/http-signature-utils/src/utils/signatures.test.ts
+++ b/packages/http-signature-utils/src/utils/signatures.test.ts
@@ -1,6 +1,5 @@
 import { generateTestKeys, TestKeys } from '../test-utils/keys'
 import { createSignatureHeaders, RequestLike } from './signatures'
-import { httpbis } from 'http-message-signatures'
 
 describe('Signature Generation', (): void => {
   let testKeys: TestKeys
--- a/packages/http-signature-utils/src/utils/signatures.test.ts
+++ b/packages/http-signature-utils/src/utils/signatures.test.ts
@@ -29,7 +29,6 @@ describe('Signature Generation', (): void => {
         return key
       })
 
-    expect(signatureParameters).toContain('alg')
     expect(signatureParameters).toContain('keyid')
     expect(signatureParameters).toContain('created')
   })
--- a/packages/http-signature-utils/src/utils/signatures.test.ts
+++ b/packages/http-signature-utils/src/utils/signatures.test.ts
@@ -1,0 +1,37 @@
+import { generateTestKeys, TestKeys } from '../test-utils/keys'
+import { createSignatureHeaders, RequestLike } from './signatures'
+import { httpbis } from 'http-message-signatures'
+
+describe('Signature Generation', (): void => {
+  let testKeys: TestKeys
+
+  beforeEach((): void => {
+    testKeys = generateTestKeys()
+  })
+
+  test('ensure that the signature input header contains the correct parameters', async (): Promise<void> => {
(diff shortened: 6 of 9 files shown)


packages/http-signature-utils/src/utils/signatures.ts
```diff
@@ -1,10 +1,9 @@
...
-  Algorithm,
-  RequestLike,
-  Signer
-} from 'http-message-signatures'
+import { type KeyLike } from 'crypto'
+import { httpbis, createSigner, Request } from 'http-message-signatures'
+
+export interface RequestLike extends Request {
+  body?: string
+}
```
raducristianpopa, line 6: The `Request` type from `http-message-signatures` does not contain the `body` property in its latest version.

packages/http-signature-utils/src/utils/signatures.ts
```diff
@@ -36,15 +28,19 @@ export const createSignatureHeaders = async ({
...
-    components,
-    parameters: {
-      created: Math.floor(Date.now() / 1000)
+
+  const signingKey = createSigner(privateKey, 'ed25519', keyId)
+
+  const { headers } = await httpbis.signMessage(
+    {
+      key: signingKey,
+      name: 'sig1',
```
raducristianpopa, line 37: In the current version, the default name for the signature is `sig`.

packages/http-signature-utils/src/utils/signatures.ts
```diff
@@ -36,15 +28,19 @@ export const createSignatureHeaders = async ({
...
+      params: ['alg', 'keyid', 'created'],
```
raducristianpopa, line 38: We need to explicitly specify the params for the signature.

packages/http-signature-utils/src/utils/signatures.ts
```diff
@@ -36,15 +28,19 @@ export const createSignatureHeaders = async ({
   if (request.body) {
     components.push('content-digest', 'content-length', 'content-type')
   }
-  const { headers } = await httpsig.sign(request, {
-    components,
-    parameters: {
-      created: Math.floor(Date.now() / 1000)
```
raducristianpopa, line 42: No need to specify `created`'s value. The library is going to automatically generate if it's passed as a parameter.

packages/http-signature-utils/src/utils/signatures.ts
mkurapov: Is it me or is there isn't any docs re: `name` and `params` attributes?

packages/http-signature-utils/src/utils/signatures.ts
```diff
@@ -17,13 +16,6 @@ export interface SignatureHeaders {
...
 }
 
-const createSigner = (privateKey: KeyLike): Signer => {
-  const signer = async (data: BinaryLike) =>
-    sign(null, data as Buffer, privateKey)
-  signer.alg = 'ed25519' as Algorithm
-  return signer
-}
-
 export const createSignatureHeaders = async ({
```
mkurapov, line 19: should we add some basic tests for this method, to show that it generates correctly named values?

packages/http-signature-utils/src/utils/signatures.ts
mkurapov: I found this commit: https://github.com/dhensby/node-http-message-signatures/commit/d23025305a970df239703b668c18477daf5e666d

> "There is no need for the body prop as signatures are not concerned with the body of the request.
All data required for signing/verifying http-signatures are based in the request line and headers."

Does this mean `body` is unnecessary in general?

I did a small test of this by updating the first tests in `validation.test.ts`, and it seems to be the case, but double check that.

packages/http-signature-utils/src/utils/signatures.ts
raducristianpopa: Documentation for the name and params attributes is missing. By looking at the type definition of `signMessage` and checking the first argument type, we can identify all the configurable options available:
```ts
export interface SignConfig extends CommonConfig {
    key: SigningKey;
    /**
     * The name to try to use for the signature
     * Default: 'sig'
     */
    name?: string;
    /**
     * The parameters to add to the signature
     * Default: see defaultParams
     */
    params?: string[];
    /**
     * The HTTP fields / derived component names to sign
     * Default: none
     */
    fields?: string[];
    /**
     * Specified parameter values to use (eg: created time, expires time, etc)
     * This can be used by consumers to override the default expiration time or explicitly opt-out
     * of adding creation time (by setting `created: null`)
     */
    paramValues?: SignatureParameters;
    /**
     * A list of supported algorithms
     */
    algs?: Algorithm[];
}
```

packages/http-signature-utils/src/utils/signatures.ts
raducristianpopa: To ensure I have a complete understanding: are the tests you are talking about, designed to verify that the `Signature-Input` includes the `alg`, `created`, and `keyId` parameters after generating?

packages/http-signature-utils/src/utils/signatures.ts
raducristianpopa: I'll double check this behaviour and follow up with my findings in a response here.

raducristianpopa: Performed some tests withing the Web Monetization extension:
1. it looks like the body is unnecessary
2. adding `content-digest`, `content-length` and `content-type` is unnecessary as well (???)

I was able to make authenticated requests with the following headers:

- Without the `Authorization` header:
```
Accept: application/json
Content-Digest: sha-512=:IuQyFymTZ4iOpIw8X3Xy873ZzY8Z9Qv1RG7ZisWfvGMLRGt4a+rsHS313udn1vzJHE9Y6W/U8jeKFh11NiWAJw==:
Content-Type: application/json
Signature: sig1=:Hp0vZAXViJOlsIivP3YdAalmE+wu/zPpRFImAxcorgtOYlEei4dghnFtJTuVQWPoRlw/oNJKim9bASjODC9oDA==:
Signature-Input: sig1=("@method" "@target-uri");alg="ed25519";keyid="4dc5df20-270e-4dc1-90f1-b26c5cff1c1a";created=1708072371
```

- With the `Authorization` header:
```
Accept: application/json
Authorization: GNAP <redacted>
Content-Digest: sha-512=:dGOTVuK0RoQ19We9JvN+ttaHKBuWMTrTFbF+PxZLtlliGgd+IaoxWsVj4THmyhBuMqIGZSiFiCdEhKBjZCpM7g==:
Content-Type: application/json
Signature: sig1=:I1oaHWnVhiPX07f9r3z9j193Ne2XNuMwHazNQ7o4OknyWbx9nDOivJL7M76j2zEWohYfGs1mavMoyj8LwPDdAg==:
Signature-Input: sig1=("@method" "@target-uri" "authorization");alg="ed25519";keyid="4dc5df20-270e-4dc1-90f1-b26c5cff1c1a";created=1708072371
```

Following the [HTTP Signatures section of the GNAP spec](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-16#section-7.3.1), it says that:
```
When the message contains a request body, the covered components MUST also include the following:

"content-digest": The Content-Digest header as defined in [[I-D.ietf-httpbis-digest-headers](https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-digest-headers-13)].
When the request message has a body, the signer MUST calculate this field value and include the 
field in the request. The verifier MUST validate this field value. REQUIRED when the message 
request contains a message body.
```

Looking at a [request example from the GNAP specification](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-16#section-7.3.1-25), I noticed it includes all `content-*` components in the `Signature-Input` header. However, I managed to make the requests successfully without including these components.

packages/http-signature-utils/src/utils/signatures.ts
mkurapov: Yup, something basic like that.

packages/http-signature-utils/src/utils/signatures.ts
mkurapov: > I was able to make authenticated requests with the following headers:

This makes sense because there is no `@content-digest` in the signature-input so it wasn't taken into account during signature generation or verification

packages/http-signature-utils/src/utils/signatures.ts
```diff
@@ -1,10 +1,9 @@
...
-  httpis as httpsig,
-  Algorithm,
-  RequestLike,
-  Signer
-} from 'http-message-signatures'
+import { type KeyLike } from 'crypto'
+import { httpbis, createSigner, Request } from 'http-message-signatures'
+
+export interface RequestLike extends Request {
+  body?: string
```
raducristianpopa, line 5: > Does this mean body is unnecessary in general?

Chose to keep the `body` property for the `RequestLike` interface because we determine in `createSignatureHeaders` whether to include `content-*` components and pass only the necessary properties when calling `signMessage`.

https://github.com/interledger/open-payments/blob/2f83578738869cf7741a34d176d9f9714772e616/packages/http-signature-utils/src/utils/signatures.ts#L34-L46

packages/http-signature-utils/src/utils/signatures.ts
```diff
@@ -36,15 +28,23 @@ export const createSignatureHeaders = async ({
...
-    parameters: {
-      created: Math.floor(Date.now() / 1000)
+
+  const signingKey = createSigner(privateKey, 'ed25519', keyId)
+
+  const { headers } = await httpbis.signMessage(
+    {
+      key: signingKey,
+      name: 'sig1',
+      params: ['alg', 'keyid', 'created'],
```
mkurapov, line 38: Interestingly:


> The explicit alg signature parameter MUST NOT be included in the signature, since the algorithm will be derived either from the key material or from the proof value.

[(source)](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-16#section-7.3.1)

raducristianpopa: This seems logical, especially considering we provide the JWK during verification, which includes the algorithm.

All tests are passing without the `alg` parameter - check https://github.com/interledger/open-payments/pull/421/commits/81eceb29748ab7ece7c11251aacbd9d1a9f7dabc. Open Payments SDK tests are failing because it expects the `alg` parameter as well.

packages/http-signature-utils/src/utils/signatures.ts
sabineschaller: > "There is no need for the body prop as signatures are not concerned with the body of the request.
All data required for signing/verifying http-signatures are based in the request line and headers."

The way I understand this is that the body itself is not required for signing but the content headers are

packages/http-signature-utils/src/utils/validation.ts
mkurapov, line 1: Based on the discussion in slack, do you think we should call `verifyContentDigest` here directly in `validateSigInputComponents`? 
I'm just wondering whether that is an http sig concern or not in general.

packages/http-signature-utils/src/utils/signatures.ts
mkurapov: In the future, I could see allowing for an arg to `createSignatureHeaders` that will take in a list of `params` that will be passed down into `signMessage`

mkurapov (approved):
Looks good, approving, just more wondering about verifying the content digest

packages/http-signature-utils/src/utils/validation.ts
raducristianpopa: Per [RFC 9421 - HTTP Message Signatures](https://www.rfc-editor.org/rfc/rfc9421.html#section-7.2.8-9):

```
Upon verification, it is important that the verifier validate not only the signature but also the value
of the Content-Digest field itself against the actual received content. Unless the verifier performs
this step, it would be possible for an attacker to substitute the message content but leave the
Content-Digest field value untouched to pass the signature. Since only the field value is covered
by the signature directly, checking only the signature is not sufficient protection against such a
substitution attack.
```

I would assume the verification should happen within the library, given its connection to HTTP Signatures. Additionally, the [GNAP spec indicates that the `Content-Digest` header verification is required](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-18#section-7.3.1-11.2.1):
```
When the message contains request content, the covered components MUST also include the following:
"content-digest":
    The Content-Digest header as defined in [[I-D.ietf-httpbis-digest-headers]. When the request message
    has content, the signer MUST calculate this field value and include the field in the request. The verifier
    MUST validate this field value. REQUIRED when the message request contains message content.

mkurapov: Want to post the HTTP message signatures blurb in the Slack thread? I think the HTP message signatures spec is stricter that the digest headers spec blurb that Nathan posted

PR #422, opened 2024-02-09
feat(op): allow custom interceptor for authenticated client

### LINKED ISSUE #412: Enable the option to provide a custom request interceptor for the authenticated client

raducristianpopa:
Allow a custom request interceptor to be passed rather than the private key and key ID. This custom interceptor will then be registered and replace the built-in interceptor for generating signatures.

## Context

Closes #412.


packages/open-payments/src/client/requests.ts
```diff
@@ -214,3 +214,30 @@ export const createAxiosInstance = (args: {
...
+        'Content-Type': 'application/json'
+      }
+    },
+    timeout: args.requestTimeoutMs
+  })
+
+  axiosInstance.interceptors.request.use(args.requestInterceptor, undefined, {
+    runWhen: (config: InternalAxiosRequestConfig) =>
+      config.method?.toLowerCase() === 'post' ||
+      !!(config.headers && config.headers['Authorization'])
```
mkurapov, line 239: I think because we are specifying POST with Authorization `runWhen` here (and not leaving it up to the caller), we should probably make the `requestInterceptor` name more explicit? like `authenticatedRequestInterceptor` or something along those lines.

.changeset/nasty-timers-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+'@interledger/open-payments': minor
```
mkurapov, line 2: Im trying to think of a better naming when it comes to the "experimental" nature of the change. Maybe I'm over-complicating this though

packages/open-payments/src/client/index.ts
```diff
@@ -258,9 +286,23 @@ export interface AuthenticatedClient
   quote: QuoteRoutes
 }
 
-export const createAuthenticatedClient = async (
+/**
+ * @experimental The `requestInterceptor` feature is currently experimental and might be removed in upcoming versions. Use at your own risk! It offers the capability to add a custom method for generating HTTP signatures. It is recommended to create the authenticated client with the `privateKey` and `keyId` arguments. If both `requestInterceptor` and `privateKey`/`keyId` are provided, an error will be thrown.
+ * @throws OpenPaymentsClientError
+ */
+export async function createAuthenticatedClient(
```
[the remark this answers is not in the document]
raducristianpopa: Even with function overloading, I think we should keep the check. While `createAuthenticatedClientArgs` will support both types without the need for type-narrowing when invoking it, the check will make sure that users fully understand the type of authenticated client they are going to use. This is mostly a runtime check, since type errors might be ignored at compile time.

packages/open-payments/src/client/index.ts
```diff
@@ -258,9 +288,37 @@ export interface AuthenticatedClient
...
+  args: CreateExperimentalAuthenticatedClientArgs
+): Promise<AuthenticatedClient>
+/**
+ * JSDoc comment here
+ * @throws OpenPaymentsClientError
+ */
+export async function createAuthenticatedClient(
```
raducristianpopa, line 302: Couldn't think of a good JSDoc comment for the default authenticated client. 🤔

packages/open-payments/src/client/index.ts
```diff
@@ -258,9 +288,37 @@ export interface AuthenticatedClient
...
+/**
+ * @experimental The `requestInterceptor` feature is currently experimental and might be removed in upcoming versions. Use at your own risk! It offers the capability to add a custom method for generating HTTP signatures. It is recommended to create the authenticated client with the `privateKey` and `keyId` arguments. If both `requestInterceptor` and `privateKey`/`keyId` are provided, an error will be thrown.
+ * @throws OpenPaymentsClientError
+ */
+export async function createAuthenticatedClient(
+  args: CreateExperimentalAuthenticatedClientArgs
+): Promise<AuthenticatedClient>
+/**
+ * JSDoc comment here
+ * @throws OpenPaymentsClientError
```
mkurapov, line 300: "Creates an Open Payments client that exposes methods to call all of the Open Payments APIs. Each request requiring authentication will be signed with the given private key."

mkurapov: Let's move the "normal" `createAuthenticatedClient` signature first. It'll be the first JSDoc autocomplete result when someone types `createAuthenticatedClient`

packages/open-payments/src/client/index.ts
```diff
@@ -239,16 +256,29 @@ export const createUnauthenticatedClient = async (
...
 
+interface InterceptorConfig {
+  /** The custom request interceptor to use. */
+  requestInterceptor: InterceptorFn
+}
+
+export type CreateAuthenticatedClientArgs = BaseAuthenticatedClientArgs &
+  PrivateKeyConfig
+
+export type CreateExperimentalAuthenticatedClientArgs =
```
mkurapov, line 279: I think the only usage of "experimental" should be in the JSdoc. Otherwise, the types should just be called what they are, eg. `CreateAuthenticatedClientWithRequestInterceptorArgs` or something like that (even though it's extremely long lol)

packages/open-payments/src/client/index.test.ts
```diff
@@ -97,5 +107,33 @@ describe('Client', (): void => {
...
+    test.each`
+      keyId          | privateKey
+      ${'my-key-id'} | ${'my-private-key'}
+      ${'my-key-id'} | ${undefined}
+      ${undefined}   | ${'my-private-key'}
+    `(
+      'throws an error if both requestInterceptor and privateKey or keyId are provided',
+      async ({ keyId, privateKey }) => {
+        try {
+          await createAuthenticatedClient({
```
mkurapov, line 120: should this have a ts-ignore tag?

raducristianpopa: Looking at it now, I expected to get a type error when running tests. However, `@swc/jest` only transpiles the code without performing any type checks.

https://swc.rs/docs/migrating-from-tsc#notes

mkurapov: That's fine, it's just to not have the code lit up red in the editor :)

packages/open-payments/src/client/requests.ts
```diff
@@ -214,3 +214,34 @@ export const createAxiosInstance = (args: {
...
+
+  axiosInstance.interceptors.request.use(
+    args.authenticatedRequestInterceptor,
+    undefined,
+    {
+      runWhen: (config: InternalAxiosRequestConfig) =>
+        config.method?.toLowerCase() === 'post' ||
+        !!(config.headers && config.headers['Authorization'])
+    }
+  )
```
mkurapov, line 244: I think the only thing that would be nice is a small test for this to check that the interceptor was added. I think you can use 
```ts
 createCustomAxiosInstance({
          requestTimeoutMs: 0,
          authenticatedRequestInterceptor: interceptor
        }).interceptors.request['handlers'][0]
```
to get a reference to the interceptor

mkurapov (commented):
Final comment

mkurapov (approved):

Issue #429, opened 2024-02-22
Update receiver regex to allow http and other fixes


BlairCurrey:
## Context
The receiver field's regex requires https and a uuid for the id. In addition, it accepts "connections" which has been deprecated for a while.

## Todos
Change receiver regex to: 
- [x] Allow http
- [x] Not require uuid format for id
- [x] Remove connection

And

- [x] Update any tests/docs as required

PR #430, opened 2024-02-22
feat: update receiver regex validation

BlairCurrey:

changes the receiver regex so that:
  -  it now matches with `http` and any id (instead of just uuid)
  -  it does not match with deprecated `/connections/` path

## Context

I noticed when using the client in the local environment that a receiver with an `http` url fails validation, which is necessary for local development. We are already using such a url, we're just not using the client for it (see the bruno/postman `Create Quote` request body). Widening that was the main goal and the other changes are more cleanup.

fixes https://github.com/interledger/open-payments/issues/429

### DIFF
--- a/openapi/schemas.yaml
+++ b/openapi/schemas.yaml
@@ -40,7 +40,7 @@ components:
     receiver:
       title: Receiver
       type: string
-      description: The URL of the incoming payment or ILP STREAM connection that is being paid.
+      description: The URL of the incoming payment that is being paid.
       format: uri
       pattern: '^(https|http)://(.+)/incoming-payments/(.+)$'
       examples:
--- a/openapi/schemas.yaml
+++ b/openapi/schemas.yaml
@@ -42,10 +42,11 @@ components:
       type: string
       description: The URL of the incoming payment or ILP STREAM connection that is being paid.
       format: uri
-      pattern: '^https://(.+)/(incoming-payments|connections)/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
+      pattern: '^(https|http)://(.+)/incoming-payments/(.+)$'
       examples:
         - 'https://ilp.rafiki.money/incoming-payments/08394f02-7b7b-45e2-b645-51d04e7c330c'
-        - 'https://ilp.rafiki.money/connections/016da9d5-c9a4-4c80-a354-86b915a04ff8'
+        - 'http://ilp.rafiki.money/incoming-payments/08394f02-7b7b-45e2-b645-51d04e7c330c'
+        - 'https://ilp.rafiki.money/incoming-payments/1'
     walletAddress:
       title: Wallet Address
       type: string
--- a/packages/openapi/src/middleware.test.ts
+++ b/packages/openapi/src/middleware.test.ts
@@ -256,6 +256,49 @@ describe('OpenAPI Validator', (): void => {
         expect(next).toHaveBeenCalled()
       }
     )
+
+    describe('Quote', (): void => {
+      test.each`
+        body                                                                                     | message                                                                              | description
+        ${{ receiver: 'ht999tp://something.com/incoming-payments' }}                             | ${'body.receiver must match pattern "^(https|http):..(.+).incoming-payments.(.+)$"'} | ${'invalid receiver, unknown protocol'}
+        ${{ receiver: 'http://something.com/incoming-payments' }}                                | ${'body.receiver must match pattern "^(https|http):..(.+).incoming-payments.(.+)$"'} | ${'invalid receiver, missing incoming payment id'}
+        ${{ receiver: 'http://something.com/connections/c3a0d182-b221-4612-a500-07ad106b5f5d' }} | ${'body.receiver must match pattern "^(https|http):..(.+).incoming-payments.(.+)$"'} | ${'invalid receiver, wrong path'}
+      `(
+        'returns 400 on invalid quote body ($description)',
+        async ({ body, message }): Promise<void> => {
+          const validateQuotePostMiddleware = createValidatorMiddleware(
+            openApi,
+            {
+              path: '/quotes',
+              method: HttpMethod.POST
+            }
+          )
+
+          const ctx = createContext(
+            {
+              headers: {
+                Accept: 'application/json',
+                'Content-Type': 'application/json'
+              }
+            },
+            {}
+          )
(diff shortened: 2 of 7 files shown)


BlairCurrey:
> ### ⚠️ No Changeset found
> Latest commit: [e10f069](https://github.com/interledger/open-payments/commit/e10f06930d8e0be3a961aa49e6d5cbe90dcffc07)
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> 
> This PR includes no changesets
> [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
> 
> [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/interledger/open-payments/new/bc/429/update-receiver-regex?filename=.changeset/four-frogs-wonder.md&value=---%0A%22%40interledger%2Fopenapi%22%3A%20patch%0A---%0A%0Afeat%3A%20update%20receiver%20regex%20validation%0A)

Note: I purposefully did not add a changeset. This only changes the schema (and a test). Can just use the fetch-schemas script in rafiki to get these changes.

openapi/schemas.yaml
```diff
@@ -42,10 +42,11 @@ components:
       type: string
       description: The URL of the incoming payment or ILP STREAM connection that is being paid.
       format: uri
-      pattern: '^https://(.+)/(incoming-payments|connections)/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
+      pattern: '^(https|http)://(.+)/incoming-payments/(.+)$'
```
BlairCurrey, line 45: As for the id change:

1. figured it was correct to loosen it to any sort of id (not necessarily uuid) after discussing with Max. Thinking was that while we use uuid's in rafiki this is really a detail which might vary from implementation to implementation and it shouldn't (I think) matter what sort of id scheme the api implementation uses. More opinions welcome.

2. I changed it to match anything but perhaps chars/numbers only would be better? I just wasn't sure about forbidding the use of query variables. While we don't support anything like this I don't know that we should have an opinion against it: `incoming-payments/123?something=1`

BlairCurrey:
> > ### ⚠️ No Changeset found
> > Latest commit: [e10f069](https://github.com/interledger/open-payments/commit/e10f06930d8e0be3a961aa49e6d5cbe90dcffc07)
> > Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> > This PR includes no changesets
> > [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
> > [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/interledger/open-payments/new/bc/429/update-receiver-regex?filename=.changeset/four-frogs-wonder.md&value=---%0A%22%40interledger%2Fopenapi%22%3A%20patch%0A---%0A%0Afeat%3A%20update%20receiver%20regex%20validation%0A)
> 
> Note: I purposefully did not add a changeset. This only changes the schema (and a test). Can just use the fetch-schemas script in rafiki to get these changes.

Looks like I was wrong about this and do need to version bump open-payments because the response validator uses the schema from this repo. I noticed while using the client in Rafiki. I manually patched the regex in `rafiki/openapi/schemas.yml` and that fixed the request validation error but I got the same error during response validation. Manually patching the schema in node_modules fixed that.

@mkurapov do you happen to know off the top of your head why pass in a spec but also refer to a different internal one like this? It just seems like we would either do one or the other.

openapi/schemas.yaml
```diff
@@ -42,10 +42,11 @@ components:
       type: string
       description: The URL of the incoming payment or ILP STREAM connection that is being paid.
```
mkurapov, line 43:
```suggestion
      description: The URL of the incoming payment that is being paid.
```

openapi/schemas.yaml
mkurapov: I think matching anything is fine in this case

mkurapov:
@BlairCurrey re:
> @mkurapov do you happen to know off the top of your head why we pass in a spec but also refer to a different internal one like this? It just seems like we would either do one or the other.

What do you mean by pass in a spec here?

BlairCurrey:
> @BlairCurrey re:
> 
> > @mkurapov do you happen to know off the top of your head why we pass in a spec but also refer to a different internal one like this? It just seems like we would either do one or the other.
> 
> What do you mean by pass in a spec here?

What I meant was passing in a spec when creating the client in rafiki, but after a more careful reading I see that is not whats happening.

I guess what I was seeing was not a validation error from the client but from the backend route's validator middleware. Which makes sense. So I take back that concern - was just confusion on my end.

packages/openapi/src/middleware.test.ts
```diff
@@ -256,6 +256,49 @@ describe('OpenAPI Validator', (): void => {
         expect(next).toHaveBeenCalled()
       }
     )
+
+    describe('Quote', (): void => {
+      test.each`
+        body                                                                                     | message                                                                              | description
+        ${{ receiver: 'ht999tp://something.com/incoming-payments' }}                             | ${'body.receiver must match pattern "^(https|http):..(.+).incoming-payments.(.+)$"'} | ${'invalid receiver, unknown protocol'}
+        ${{ receiver: 'http://something.com/incoming-payments' }}                                | ${'body.receiver must match pattern "^(https|http):..(.+).incoming-payments.(.+)$"'} | ${'invalid receiver, missing incoming payment id'}
+        ${{ receiver: 'http://something.com/connections/c3a0d182-b221-4612-a500-07ad106b5f5d' }} | ${'body.receiver must match pattern "^(https|http):..(.+).incoming-payments.(.+)$"'} | ${'invalid receiver, wrong path'}
```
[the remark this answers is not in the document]
BlairCurrey: I'll make some happy-path tests in a new test group.

Im not necessarily against hardcoding the message here but just want to note that the test is for the quote resource in general and could be extended to validate different fields which would require parameterizing the message like this,  like the existing tests do for `/incoming-payments`.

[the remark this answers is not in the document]
BlairCurrey: removed message param and added happy path cases

mkurapov (approved):
LGTM but make sure to get someone else's eyes on this as well :+1:

njlie (approved):

PR #434, opened 2024-03-08
fix: make interact_ref non required

### LINKED ISSUE #433: Make `interact_ref` optional on POST `/continue/{id}`
## Context
The `interact_ref` is not actually required in the body of continuation requests. In the case of a grant request with no `interact.finish` method, we should be able to poll the continuation endpoint without an `interact_ref`. But it's still required by the schema/client.

## Todos
- [ ] Update spec to make interact_ref optional on POST `/continue/{id}`
- [ ] Ensure client can make continuation requests without an `interact_ref`
- [ ] update bruno collection to use this method and update the docs https://github.com/interledger/rafiki/pull/2507#discussion_r1516135752

sabineschaller:

Apparently, my tool did some formatting changes, too. Sorry!

## Context

- fixes #433

### DIFF
--- a/.changeset/red-bulldogs-type.md
+++ b/.changeset/red-bulldogs-type.md
@@ -1,0 +1,5 @@
+---
+'@interledger/open-payments': patch
+---
+
+updating types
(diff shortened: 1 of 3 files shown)


mkurapov (approved):
Were you able to test this in the local playground open payments example?

sabineschaller:
> Were you able to test this in the local playground open payments example?

Can I? Do I not have to fetch the published version of this first and hence, need to publish it first?

PR #436, opened 2024-03-09
feat: make body, interact_ref optional on client grant continue

BlairCurrey:

Updates the client args so that the continue request does not require the `interact_ref` nor `body`. It was recently changed such that it can be undefined, but is still required like so:

```ts
client.grant.continue({ accessToken, url }, { interact_ref: undefined })
```

With these changes you can do:

```ts
client.grant.continue({ accessToken, url }, {})
```

or

```ts
client.grant.continue({ accessToken, url })
```

## Context

small follow-up on https://github.com/interledger/open-payments/issues/433

Originally I made the PR for ergonomics sake on the client, but actually I think it will be necessary for continuation polling via the client (at least without adjusting auth server logic), since this is how we're checking whether or not to poll (condition wont be met with any `body.interact_ref`):

```ts
  if (!ctx.request.body || Object.keys(ctx.request.body).length === 0) {
    await pollGrantContinuation(deps, ctx, continueId, continueToken)
    return
  }
```

### DIFF
--- a/.changeset/olive-eagles-talk.md
+++ b/.changeset/olive-eagles-talk.md
@@ -1,0 +1,5 @@
+---
+'@interledger/open-payments': patch
+---
+
+Makes body and interact_ref optional on continuation route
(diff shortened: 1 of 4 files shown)


sabineschaller (approved):
Thank you for fixing my sloppy work!

PR #440, opened 2024-03-14
439/mk/export openapi specs

### LINKED ISSUE #439: Export functions that allow fetching the Open Payments OpenAPI objects in the Open Payments client
## Context
Since we pack and ship the yaml files directly with the `open-payments` package, we should have functionality that exports out the OpenAPI objects generated from the Open Payments OpenAPI yaml specs. 

- This is primarily to aid with https://github.com/interledger/rafiki/issues/1562, as this will enable `backend` `auth` packages in Rafiki to call functions, and use the OpenAPI validation functions for the Open Payment specs without needing to store the yaml spec files locally.

## Todos
- [ ] Export functions allowing use of OpenAPI objects directly

mkurapov:

- Export `getAuthServerOpenAPI` `getResourceServerOpenAPI` `getWalletAddressServerOpenAPI` functions from the `open-payments` package
- Move `openapi/*yaml` files into `openapi/specs/`
 
 

## Context
Since we bundle the yaml files with the `open-payments` package, we should let users of the package (importantly for us, Rafiki) to fetch the `OpenAPI` objects for the Open Payments specs

Fixes #439


BlairCurrey (approved):
Looks good to me. Will be nice to have an easier way to get the right spec version in rafiki.

I spent some time thinking about the path assertions in `packages/open-payments/src/openapi/index.test.ts`. I was just thinking they were more a concern of `createOpenAPI` (if anything - kinda just an implementation detail of the spec) but I suppose it makes sense as a sanity check and isn't hard to update when things change (and they dont change often).

mkurapov:
@BlairCurrey also, what was cool is that I just pnpm-linked the package with the changes and ran it through your integration test as a sanity in Rafiki :+1:

PR #446, opened 2024-03-25
feat(openapi): throw OpenAPIValidatorMiddlewareError

### LINKED ISSUE #445: OpenAPI validator middleware should throw a custom error
## Context
While working on 
- https://github.com/interledger/rafiki/issues/1905

in Rafiki, I discovered that the OpenAPI validator middleware uses the `ctx.throw` method directly when handling errors:
https://github.com/interledger/open-payments/blob/d365729e5a3df7127f82071b31185490c2e9d9ef/packages/openapi/src/middleware.ts#L29-L38 

If we want to use a custom error (like the planned `OpenPaymentsServer` error in Rafiki) this prevents us from writing a [general error error handler](https://github.com/koajs/koa/blob/master/docs/error-handling.md#default-error-handler) at the beginning of the middleware chain, since this middleware will always throw `ctx.throw(500)` for something other than a Koa.HttpError (and have a busy `console.error` log as well, without control over it). 

By throwing a custom error instead of a `ctx.throw`, we can directly determine whether the particular Koa error came from this validation middleware or not, allowing us to be more explicit in how we handle the error and how we log the contents of the error.

## Todos
- [ ] Have the OpenAPI middleware throw a custom error (e.g. `OpenAPIValidatorMiddlewareError`)

mkurapov:
Changes to the `createValidatorMiddleware` behaviour in `packages/openapi`:

- The middleware now throws a `OpenAPIValidatorMiddlewareError` instead of directly using Koa's `ctx.throw` function. This error class has `message` and `status` properties that describe the exact validation error as well as the status code that should be thrown.
- The middleware now takes in an optional `validationOptions` argument that determines whether to validate just the request, the response or both. By default, both are validated, and the middleware no longer uses `process.env.NODE_ENV` to determine whether to validate the response or not.

## Context

Fixes #445

### DIFF
--- a/.changeset/tasty-spoons-pay.md
+++ b/.changeset/tasty-spoons-pay.md
@@ -1,0 +1,8 @@
+---
+'@interledger/openapi': major
+---
+
+Changes to the `createValidatorMiddleware` behaviour:
+
+- The middleware now throws a `OpenAPIValidatorMiddlewareError` instead of directly using Koa's `ctx.throw` function. This error class has `message` and `status` properties that describe the exact validation error as well as the status code that should be thrown.
+- The middleware now takes in an optional `validationOptions` argument that determines whether to validate just the request, the response or both. By default, both are validated, and the middleware no longer uses `process.env.NODE_ENV` to determine whether to validate the response or not.
--- a/packages/openapi/src/index.ts
+++ b/packages/openapi/src/index.ts
@@ -155,10 +155,14 @@ class OpenAPIImpl implements OpenAPI {
   }
 }
 
+interface TransformedError {
+  message: string
+}
+
 const errorTransformer = (
   _openapiError: OpenAPIResponseValidatorError,
   ajvError: ErrorObject
-) => {
+): TransformedError => {
   // Remove preceding 'data/'
   // Delineate subfields with '.'
   const message = ajv.errorsText([ajvError]).slice(5).replace(/\//g, '.')
@@ -184,7 +188,7 @@ const customFormats = {
 
 export interface ValidationError {
   status?: number
-  errors: ReturnType<typeof errorTransformer>[]
+  errors: TransformedError[]
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/packages/openapi/src/index.ts
+++ b/packages/openapi/src/index.ts
@@ -182,7 +182,7 @@ const customFormats = {
   }
 }
 
-interface ValidationError {
+export interface ValidationError {
   status?: number
   errors: ReturnType<typeof errorTransformer>[]
 }
--- a/packages/openapi/src/index.ts
+++ b/packages/openapi/src/index.ts
@@ -184,7 +184,7 @@ const customFormats = {
 
 interface ValidationError {
   status?: number
-  errors: string[]
+  errors: ReturnType<typeof errorTransformer>[]
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
--- a/packages/openapi/src/middleware.test.ts
+++ b/packages/openapi/src/middleware.test.ts
@@ -2,7 +2,13 @@ import Koa from 'koa'
 import * as httpMocks from 'node-mocks-http'
 import { v4 as uuid } from 'uuid'
 import assert from 'assert'
-import { createOpenAPI, OpenAPI, HttpMethod } from './'
+import {
+  createOpenAPI,
+  OpenAPI,
+  HttpMethod,
+  RequestValidator,
+  ResponseValidator
+} from './'
 import {
   createValidatorMiddleware,
   OpenAPIValidatorMiddlewareError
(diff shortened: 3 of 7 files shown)


.changeset/tasty-spoons-pay.md
```diff
@@ -0,0 +1,8 @@
+---
+'@interledger/openapi': major
```
mkurapov, line 2: I _think_ it should be major since it changes how `createValidatorMiddleware` is used.
But don't mind using `minor` since its just Rafiki using it

packages/openapi/src/middleware.ts
```diff
@@ -2,39 +2,76 @@ import { OpenAPI, RequestOptions, isValidationError } from './'
...
+  options: RequestOptions,
+  validationOptions: ValidationOptions | undefined = {
+    validateRequest: true,
+    validateResponse: true
+  }
 ): (ctx: Koa.Context, next: () => Promise<unknown>) => Promise<void> {
-  const validateRequest = spec.createRequestValidator<T['request']>(options)
-  const validateResponse =
-    process.env.NODE_ENV !== 'production' &&
-    spec.createResponseValidator(options)
```
mkurapov, line 12: I feel like directly using `process.env.NODE_ENV` is not a fun time.
Will update Rafiki to have `validationOptions.validateReponse: process.env.NODE_ENV !== 'production'` instead

packages/openapi/src/middleware.ts
```diff
@@ -2,39 +2,76 @@ import { OpenAPI, RequestOptions, isValidationError } from './'
...
+      if (validationOptions?.validateResponse) {
+        responseValidator(ctx.response)
       }
     } catch (err) {
-      if (err instanceof Koa.HttpError) {
-        throw err
-      } else if (isValidationError(err)) {
-        ctx.throw(err.status ?? 500, err.errors[0])
-      } else {
-        console.log('err=', err)
```
mkurapov, line 35: removing pesky console.log here

.changeset/tasty-spoons-pay.md
BlairCurrey: Major sounds right, if I'm understanding correctly. This will break consumers where `NODE_ENV` is `production` and a response will fail validation.

packages/openapi/src/index.ts
```diff
@@ -182,9 +182,9 @@ const customFormats = {
   }
 }
 
-interface ValidationError {
+export interface ValidationError {
   status?: number
-  errors: string[]
+  errors: ReturnType<typeof errorTransformer>[]
```
BlairCurrey, line 187: The return type of `errorTransformer` is 

```ts
{
  message: string
}
```

Might it make sense to define that and use it here and the `errorTransformer` signature?

packages/openapi/src/middleware.ts
```diff
@@ -2,39 +2,74 @@ import { OpenAPI, RequestOptions, isValidationError } from './'
...
-        ctx.throw(500)
+      if (isValidationError(err)) {
+        throw new OpenAPIValidatorMiddlewareError(
+          `Received error validating OpenAPI request: ${err.errors[0]?.message}`,
+          err.status || 500
+        )
       }
+
+      throw err // Should not be possible (only ValidationError is thrown in requestValidator)
     }
```
BlairCurrey, line 46: I think this is functionally equivalent but perhaps a bit more explicit? only try/catching `requestValidator` basically.

```ts
    if (validationOptions?.validateRequest) {
      if (!ctx.accepts('application/json')) {
        throw new OpenAPIValidatorMiddlewareError(
          'Received error validating OpenAPI request: Must accept application/json',
          406
        )
      }
      try {
        requestValidator(ctx.request)
      } catch (err) {
        if (isValidationError(err)) {
          throw new OpenAPIValidatorMiddlewareError(
            `Received error validating OpenAPI request: ${err.errors[0]?.message}`,
            err.status || 500
          )
        }

        throw err // Should not be possible (only ValidationError is thrown in requestValidator)
      }
    }
```

As-is, the 406 `OpenAPIValidatorMiddlewareError` is being re-thrown where the comment suggests it should be impossible. I guess in that case we could just remove the comment, but it seems like catching it is unnecessary in the first place.

packages/openapi/src/middleware.test.ts
BlairCurrey, line 1: Can we test some of the non-default options on `createValidatorMiddleware`? I guess maybe mocking the return of `spec.createRequestValidator` (and response) and verify they arent called when false?

packages/openapi/src/middleware.ts
mkurapov: Yup, definitely good call. Updating.

packages/openapi/src/index.ts
mkurapov: Yup, I think I went with this since I didn't have a good name for the type 😆

BlairCurrey: Notice that I didn't offer a name myself lol

BlairCurrey (approved):

Issue #415, opened 2024-02-05
Make quoting optional in the resource server spec


mkurapov:
## Context
During the WM work week it was decided to make quoting optional so that the WM flow would be less chatty. This requires changing how outgoing payments are created.

We need to change the OP spec for outgoing payment creation to allow to take in an object with an `incomingPaymentId` as well as a `debitAmount` (instead of only having the option of passing in a `quoteId`)

## Todos
- [ ] Change resource server spec to take in a `incomingPaymentId` as well as a `debitAmount` (in the cases where `quoteId` is not provided)

PR #455, opened 2024-04-02
feat: add option to create outgoing payment from incoming payment

BlairCurrey:
- changes spec and test to allow createing outgoing payment  from `incomingPaymentId`/`debitAmount` instead of only `quoteId`

## Context

fixes https://github.com/interledger/open-payments/issues/415

### DIFF
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,1 +1,0 @@
-singleQuote: true
--- a/docs/src/content/docs/introduction/op-concepts.mdx
+++ b/docs/src/content/docs/introduction/op-concepts.mdx
@@ -90,12 +90,7 @@ When an `outgoing-payment` is completed against an open and active `incoming-pay
 
 To set up ILP as a payment method in OP, the following are required:
 
-- The <LinkOut href='https://interledger.org/developers/rfcs/ilp-addresses/'>ILP address
-  </LinkOut> of the payee’s ASE: The ILP address is required so that packets representing
-payments routed over the Interledger network will be forwarded to the node owned
-and operated by the intended receiver (i.e. payee's ASE). The `ilpAddress` field
-must be sent by the payee’s ASE in the `methods` object on an `incoming-payment`
-response.
+- The <LinkOut href='https://interledger.org/developers/rfcs/ilp-addresses/'>ILP address </LinkOut> of the payee’s ASE: The ILP address is required so that packets representing payments routed over the Interledger network will be forwarded to the node owned and operated by the intended receiver (i.e. payee's ASE). The `ilpAddress` field must be sent by the payee’s ASE in the `methods` object on an `incoming-payment` response.
 - Shared secret: A cryptographically secured secret exchanged between the sender and receiver to ensure that packets sent over the Interledger network through a <LinkOut href='https://interledger.org/developers/rfcs/stream-protocol/'>STREAM</LinkOut> connection can only be read by the two parties. The payee’s ASE must return the `sharedSecret` field in the `methods` object on an `incoming-payment` response.
 - The payee's ASE must return the value `ilp` in the `type` field of the `methods` object on an `incoming-payment` response.
 - The payer's ASE must send the value `ilp` in the `method` field when creating a `quote`.
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -291,27 +291,60 @@ paths:
                   quoteId: 'https://ilp.rafiki.money/quotes/ab03296b-0c8b-4776-b94e-7ee27d868d4d'
                   metadata:
                     externalRef: INV2022-02-0137
+              Create an outgoing payment based on an incoming payment:
+                value:
+                  walletAddress: 'https://ilp.rafiki.money/alice/'
+                  incomingPaymentId: 'https://ilp.rafiki.money/incoming-payments/8d4e4776-2e55-4e5a-bcbe-8348ed1e86de'
+                  debitAmount:
+                    value: '2500'
+                    assetCode: USD
+                    assetScale: 2
+                  metadata:
+                    externalRef: INV2022-02-0137
             schema:
-              type: object
-              properties:
-                walletAddress:
-                  $ref: ./schemas.yaml#/components/schemas/walletAddress
-                quoteId:
-                  type: string
-                  format: uri
-                  description: The URL of the quote defining this payment's amounts.
-                metadata:
-                  type: object
-                  additionalProperties: true
-                  description: Additional metadata associated with the outgoing payment. (Optional)
-              required:
-                - quoteId
-                - walletAddress
-              additionalProperties: false
+              oneOf:
+                - type: object
+                  properties:
+                    walletAddress:
+                      $ref: ./schemas.yaml#/components/schemas/walletAddress
+                    quoteId:
+                      type: string
+                      format: uri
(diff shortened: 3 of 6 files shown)


packages/open-payments/src/client/outgoing-payment.test.ts
```diff
@@ -280,38 +281,58 @@ describe('outgoing-payment', (): void => {
...
-      ${{ description: 'Some description', externalRef: '#INV-1' }}
-      ${undefined}
-    `('creates outgoing payment', async ({ metadata }): Promise<void> => {
-      const outgoingPayment = mockOutgoingPayment({
+      quoteId      | incomingPaymentId    | debitAmount    | metadata                                                      | createSource
+      ${quoteId_}  | ${undefined}         | ${undefined}   | ${{ description: 'Some description', externalRef: '#INV-1' }} | ${'quote'}
+      ${quoteId_}  | ${undefined}         | ${undefined}   | ${undefined}                                                  | ${'quote'}
+      ${undefined} | ${incomingPaymentId} | ${debitAmount} | ${undefined}                                                  | ${'incoming payment'}
+    `(
+      'creates outgoing payment from $createSource',
```
BlairCurrey, line 298: refactors our create outgoing payment test to not assume `quoteId`. Now either `quoteId` or `incomingPaymentId`/`debitAmount` are passed into the test as parameters.

BlairCurrey:
converted to draft while i clarify some requirements (and most likely) update a few things to meet them.

openapi/resource-server.yaml
```diff
@@ -291,27 +291,60 @@ paths:
                   quoteId: 'https://ilp.rafiki.money/quotes/ab03296b-0c8b-4776-b94e-7ee27d868d4d'
                   metadata:
                     externalRef: INV2022-02-0137
+              Create an outgoing payment based on an incoming payment:
+                value:
+                  walletAddress: 'https://ilp.rafiki.money/alice/'
+                  incomingPaymentId: 'https://ilp.rafiki.money/incoming-payments/8d4e4776-2e55-4e5a-bcbe-8348ed1e86de'
```
sabineschaller, line 297: can we just call that `incomingPayment` without the `Id`-part? I think that is confusing.

openapi/resource-server.yaml
```diff
@@ -291,27 +291,60 @@ paths:
...
-              required:
-                - quoteId
-                - walletAddress
-              additionalProperties: false
+              oneOf:
+                - type: object
+                  properties:
+                    walletAddress:
+                      $ref: ./schemas.yaml#/components/schemas/walletAddress
+                    quoteId:
```
sabineschaller, line 310: I think in that vain, we should probably also rename that to just be `quote`?

sabineschaller (commented):
Just some renaming suggestions.

openapi/resource-server.yaml
BlairCurrey: I have always found this confusing as well (I would assume id is the uuid) but I think it's the most consistent, and I think `incomingPayment` would also be confusing. 

Looking at the rest of the spec ids are generally referring to urls. such as `id` on the `incomingPayment` object and `quoteId` on the outgoing payment object.

Personally I would opt for `incomnigPaymentUrl` or `incomnigPaymentUri` as `incomingPayment` seems like it would be the full object (which itself has an id which is a url). I guess a counterpoint is the `walletAddress` which i see is used for both a url and an object. Although I would argue that shouldn't be emulated.

BlairCurrey: still mulling this over...

sabineschaller: Good point, `...Url` or `...Uri` is probably the best way to name it. It's a bit breaking change though... Maybe we keep it at `...Id` for now and bring it up during the Open Payments call

njlie: I think that  `incomingPayment` without any specifiers like `id` in the name is pretty open to interpretation, so long as we're consistent. [Stripe mixes it up between a fully expanded object or just an id between fields in the same object](https://docs.stripe.com/api/charges/object), for example. So in our case I think we could have it be a URL if we wanted.

mkurapov: :+1: for `-Url`

BlairCurrey: In our planning call we decided on `incomingPayment` because its clearer and no less consistent than `incomingPaymentId`. Doing `incomingPaymentUrl` would be a new pattern and would require substantial renaming. Will discuss in the Open Payments call and tackle that in a separate issue, if any.

openapi/resource-server.yaml
BlairCurrey: I was going to rename this to `quote` to make this endpoint consistent internally but this will be harder than I initially realized. We are currently tightly coupled to the [open payments snippets repo](https://github.com/interledger/open-payments-snippets) via [these links in our `<ChunkedSnippets>` in the docs](https://github.com/interledger/open-payments/blob/main/docs/src/content/docs/guides/make-recurring-payments.mdx?plain=1#L168-L180).

This means if we rename to `quote`, the docs will be wrong until we update open payments snippets. Not great really. I think we have two options:

- not rely on the open payments snippets in our docs 

OR

- if we want to depend on the open payments snippets, make the snippets a package in this repository

Just to be clear, the snippets reflect whatever version of the open-payments package it has as a dependency. That can be behind the latest. That's fine. It's those references in the docs which assume the Open Payments snippets are up to date with the latest changes here which is the problem.

BlairCurrey: To be clear I don't intend on tackling any of this here - just explaining why I opted not to rename to `quote`. If we make any changes along these lines we can do so in another task.

.prettierrc.yaml
BlairCurrey, line 1: it's already defined in package.json (along with more rules)

docs/src/content/docs/introduction/op-concepts.mdx
```diff
@@ -90,12 +90,7 @@ When an `outgoing-payment` is completed against an open and active `incoming-pay
 
 To set up ILP as a payment method in OP, the following are required:
 
-- The <LinkOut href='https://interledger.org/developers/rfcs/ilp-addresses/'>ILP address
-  </LinkOut> of the payee’s ASE: The ILP address is required so that packets representing
-payments routed over the Interledger network will be forwarded to the node owned
-and operated by the intended receiver (i.e. payee's ASE). The `ilpAddress` field
-must be sent by the payee’s ASE in the `methods` object on an `incoming-payment`
-response.
+- The <LinkOut href='https://interledger.org/developers/rfcs/ilp-addresses/'>ILP address </LinkOut> of the payee’s ASE: The ILP address is required so that packets representing payments routed over the Interledger network will be forwarded to the node owned and operated by the intended receiver (i.e. payee's ASE). The `ilpAddress` field must be sent by the payee’s ASE in the `methods` object on an `incoming-payment` response.
```
BlairCurrey, line 93: Just fixes a prettier error: https://github.com/interledger/open-payments/actions/runs/8710067273/job/23891288640

Kind of weird - it failed here before I merged in main. But worked locally. Then when I merged main in it started failing locally. So a couple mysteries:

1. how did it get in main in the first place? should have failed in ci on branch
2. why did my pr fail before I had main merged in?

sabineschaller (approved):
Thank you! Let's resolve that whole `id` mess in a subsequent PR.

Issue #454, opened 2024-04-02
Prepare @interledger/open-payments lib for use with manifest V3


mkurapov:
## Context
There are a few blockers when trying to use the @interledger/open-payments in the Web Monetization extension with the V3 manifest. The following should be investigated/completed before we fan support the library in the WM V3 extension.

## Todos

- [ ] Replace axios in the library since `XMLHttpRequest` is unavailable in service workers (that the extension uses), which Axios relies on:

https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest_API
https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers#replace-xmlhttprequest

Possible solutions:
* ✅ using something like [ky](https://github.com/sindresorhus/ky) (http client built on top of fetch with hooks - beforeRequest, afterRequest …)  _(ky is the selected option as it provides a good API for hooks that allow us to sign requests in the SDK before making API calls)_

* the option to pass an adapter when initializing the clients https://www.npmjs.com/package/@vespaiach/axios-fetch-adapter
* move to fetch - we will have to write our own interceptors

- [ ] Make schema validation optional in the SDK. This is for the following reasons:

### Failed solution for schema validation:

- [ ] Update the `@interledger/openapi` library to not need to use `$Ref` parser library at runtime:

> The $Ref parser library is using @jsdevtools/ono which is calling eval [under the hood](https://github.com/search?q=repo%3AJS-DevTools%2Fono%20eval&type=code). For Manifest V2, I had to add the unsafe-eval option for the extension CSP (which we shouldn’t really ship with) and everything is working fine. If this option is removed we would not be able to load the schemas.
For Manifest V3, this is not going to be that straight-forward. eval [is not allowed in Manifest V3](https://developer.chrome.com/docs/extensions/reference/manifest/content-security-policy#extension_pages_policy):
The extension_pages policy cannot be relaxed beyond this minimum value. In other words, you cannot add other script sources to directives, such as adding 'unsafe-eval' to script-src. If you add a disallowed source to your extension's policy, Chrome will throw an error like this at install time

> The $Ref parser can not resolve the schema path:
When moving to MV3, the library can not load the YAML schemas:
[...]

PR #461, opened 2024-04-18
feat(open-payments): replace axios with ky

mkurapov:

First part of making the open payments SDK usable in the web extension manifest V3:
Replace axios with a wrapper for native fetch ([ky](https://github.com/sindresorhus/ky)). This is due to the fact that in the web extension we cannot use `XMLHttpRequest`  that axios uses, but native fetch is (of course) supported.

## Context

Progress toward #454


jest.config.base.js
```diff
@@ -2,7 +2,7 @@
 
 module.exports = {
   transform: {
-    '^.+\\.tsx?$': ['@swc/jest']
+    '^.+\\.(t|j)sx?$': ['@swc/jest']
```
mkurapov, line 5: so we transpile JS files as well, particularly for the possible files in `transformIgnorePatterns` -> `node_modules/ky`

packages/open-payments/package.json
```diff
@@ -32,16 +32,16 @@
...
     "@types/uuid": "^9.0.0",
-    "nock": "^13.3.0",
+    "nock": "14.0.0-beta.5",
     "openapi-typescript": "^4.5.0",
     "typescript": "^4.9.5"
   },
   "dependencies": {
     "@interledger/http-signature-utils": "workspace:2.0.2",
     "@interledger/openapi": "workspace:2.0.1",
-    "axios": "^1.6.0",
```
mkurapov, line 42: 👋

packages/open-payments/package.json
```diff
@@ -32,16 +32,16 @@
   "devDependencies": {
     "@types/node": "^20.12.7",
     "@types/uuid": "^9.0.0",
-    "nock": "^13.3.0",
+    "nock": "14.0.0-beta.5",
```
mkurapov, line 35: this is the [version](https://github.com/nock/nock) with experimental support for fetch

packages/open-payments/jest.config.js
```diff
@@ -13,5 +15,6 @@ module.exports = {
   modulePaths: [`<rootDir>/packages/${packageName}/src/`],
   id: packageName,
   displayName: packageName,
-  rootDir: '../..'
+  rootDir: '../..',
+  transformIgnorePatterns: [`node_modules/(?!.pnpm|${esModules.join('|')})`]
```
mkurapov, line 19: transpile ESM ky package into CommonJS

packages/open-payments/src/client/requests.ts
```diff
@@ -158,90 +199,113 @@ const checkUrlProtocol = (deps: BaseDeps, url: string): string => {
...
+)
+
+export type HttpClient = KyInstance
+
+export type InterceptorFn = (request: Request) => Request | Promise<Request>
+
+export const createHttpClient = async (
+  args: CreateHttpClientArgs
+): Promise<HttpClient> => {
+  const { default: ky } = await import('ky')
```
mkurapov, line 216: ky is ESM only

packages/open-payments/src/client/requests.ts
```diff
@@ -111,25 +129,45 @@ export const deleteRequest = async <TResponse>(
     })
 
     openApiResponseValidator({
-      status,
-      body: data || undefined
+      status: response.status,
+      body: undefined
```
mkurapov, line 133: we always expect 204 in our requests which always come w/o body

tsconfig.build.json
```diff
@@ -43,7 +43,7 @@
     // "noPropertyAccessFromIndexSignature": true,  /* Require undeclared properties from index signatures to use element accesses. */
 
     /* Module Resolution Options */
-    "moduleResolution": "node" /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */,
+    "moduleResolution": "NodeNext" /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */,
```
mkurapov, line 46: this is so we _don't_ transpile dynamic import (the one that we use for ky) back to `require`: https://www.typescriptlang.org/tsconfig/#moduleResolution

.changeset/lemon-gifts-hide.md
```diff
@@ -0,0 +1,5 @@
+---
+'@interledger/open-payments': minor
```
BlairCurrey, line 2: should be major since this is breaking: https://github.com/interledger/open-payments/pull/461/files#diff-2d8be9fd3b70cc62b8a63016a7bf60f5d21f191b9e238966a0ed7170957a77d6R36

packages/open-payments/src/client/incoming-payment.test.ts
```diff
@@ -64,7 +66,7 @@ describe('incoming-payment', (): void => {
         },
         openApiValidators.successfulValidator
       )
-      expect(result).toStrictEqual(incomingPayment)
+      expect(result).toEqual(incomingPayment)
```
BlairCurrey, line 69: Can you elaborate on why this was necessary? I changed back to `toStrictEqual` and it fails with `serializes to the same string`. Something that changed with nock? or?

packages/open-payments/src/client/requests.ts
```diff
@@ -140,10 +178,13 @@ const handleError = (
   }
 
   const errorMessage = `Error making Open Payments ${requestType} request`
-  deps.logger.error({ status: errorStatus, errorDescription }, errorMessage)
+  deps.logger.error(
+    { status: errorStatus, errorDescription, url, requestType },
+    errorMessage
+  )
```
BlairCurrey, line 184: This is probably quite the edge case, but `errorStatus` and `errorDescription` could be undefined if the error is not a validation error, and `HTTPError`, or `Error`. Just wondering if we can handle that better such as logging the original error or something.

packages/open-payments/src/client/requests.ts
```diff
@@ -158,90 +199,113 @@ const checkUrlProtocol = (deps: BaseDeps, url: string): string => {
...
+
+export type InterceptorFn = (request: Request) => Request | Promise<Request>
+
+export const createHttpClient = async (
+  args: CreateHttpClientArgs
+): Promise<HttpClient> => {
+  const { default: ky } = await import('ky')
+
+  const kyInstance = ky.create({
+    timeout: args?.requestTimeoutMs,
```
BlairCurrey, line 219:
```suggestion
    timeout: args.requestTimeoutMs,
```

no optional chaining necessary as far as I can tell

packages/open-payments/src/client/requests.ts
```diff
@@ -158,90 +199,113 @@ const checkUrlProtocol = (deps: BaseDeps, url: string): string => {
...
+  if ('authenticatedRequestInterceptor' in args) {
+    requestInterceptor = (request) => {
+      if (requestShouldBeAuthorized(request)) {
+        return args.authenticatedRequestInterceptor(request)
       }
-    )
+
+      return request
+    }
+  } else if ('privateKey' in args && 'keyId' in args) {
```
BlairCurrey, line 236: arguably a code style thing, but this could just be `else` and typescript will know args has `privateKey` and `keyId`. Not necessarily against the explicitness of this but it also kinda suggests args could have neither `'authenticatedRequestInterceptor'` or `'privateKey'` and `'keyId'`.

BlairCurrey, line 236: feel free to disregard if you like it as-is, but figured i'd leave this suggestion for convenience

```suggestion
  } else {
```

packages/open-payments/jest.config.js
mkurapov: I think what will end up working is doing the same in Rafiki and by adding the Adonis library to the list of esmModules. Haven't yet to try, but it follows the same principle

packages/open-payments/src/client/incoming-payment.test.ts
mkurapov: I thought this was odd as well, but I think it's the difference between how a response is returned in axios vs fetch. It's possible the order in which the object keys are returned are different?

Someone has an unaswered question related exactly to this: https://stackoverflow.com/questions/76211440/why-does-jests-tostrictequal-fail-to-match-json-responses-made-using-the-fetch

packages/open-payments/src/client/requests.ts
mkurapov: Yes, thanks. Updating

packages/open-payments/src/client/requests.ts
mkurapov: Makes sense, we also do a check at runtime when initializing the client anyway :+1:

BlairCurrey (commented):
Beyond the couple suggestions and questions I left, it looks good with the caveat that I couldn't get it to work when plugging this into Rafiki's integration tests. Although perhaps you have more details or already figured it out. When running it I got:

```
console.error
    TypeError: A dynamic import callback was invoked without --experimental-vm-modules
```

You mentioned the jest config would need the same `transformIgnorePatterns` and `transform` settings as here but after adding those I still see an error:

```
ReferenceError: esModules is not defined
```

mkurapov:
(Spoke in slack, but adding for visibility)

is how we can support this change of the open payments library in rafiki tests: https://github.com/interledger/rafiki/compare/main...mk/op-client-ky-tests

BlairCurrey (approved):

packages/open-payments/src/client/requests.ts
mkurapov: Yep, seems quite unlikely, but updated to log in any case :+1:

raducristianpopa (approved):

PR #467, opened 2024-04-29
feat(open-payments): make schema validation in the open-payments client behind a flag

### LINKED ISSUE #454: Prepare @interledger/open-payments lib for use with manifest V3

mkurapov:
## Context
This change adds `validateResponses` flag to open payments client initialization function. This allows bypassing runtime response validation against the Open Payments OpenAPI specs. 

This is particularly useful for the WM extension use case since the openapi validator libraries we use under the hood in @interledger/openapi use `ajv` schema validation during runtime, which uses unsafe `eval` , and as a result, prevents us from running it in the extension for manifest v3: https://github.com/ajv-validator/ajv/issues/406.

Fixes #454


packages/open-payments/src/client/index.ts
```diff
@@ -129,6 +130,7 @@ const parseKey = (
 
 const createUnauthenticatedDeps = async ({
   useHttp = false,
+  validateResponses = true,
```
mkurapov, line 133: the main change here: the addition of optional `validateResponses` flag when creating the open payments clients.

BlairCurrey (approved):
I think you're waiting on Radu for a test in the browser, but it looks good to me

mkurapov:
Tested in the browser with the extension, I do not see any errors related to the client initialization.

Issue #469, opened 2024-04-30
Return the remaining balance on the outgoing payment grant during outgoing payment creation


mkurapov:
## Context
For the web monetization extension (and very likely other use cases), we should know how much has already been spent on the outgoing payment grant. 

Whenever we create an outgoing payment, we determine whether the amount for the payment fits within the limits of the issued outgoing payment grant (by summing up all of the other payment that have been created under that said grant). If not, we reject the payment, otherwise, we create it. This accounting is done in the Open Payment resource server because that is where all of the outgoing payments are stored. We can return these amounts as part of the outgoing payment creation response.

## Todos
- [ ] Update the RS OpenAPI spec to return `grantSpentDebitAmount: Amount` and `grantSpentReceiveAmount: Amount` in the outgoing payment creation response

Issue #471, opened 2024-05-08
Update how we set limits during outgoing payment grant creation


mkurapov:
## Context
Currently, when we create an outgoing payment grant request, we can pass in a `limits` attribute with debitAmount and/or receiveAmount. For simplicity of the API as well as the simplicity of documenting how these amounts are used, we should update the spec to allow only setting `limits.debitAmount` OR `limits.receiveAmount`, not both. 

## Todos
- [ ] Update the auth-server spec to allow debitAmount or receiveAmount in the `limits` attribute of outgoing payment grant creation

PR #474, opened 2024-05-20
feat: return outgoing-payment-with-spent-amounts from POST /outgoing-payment

BlairCurrey:

- changes `POST /outgoing-payment` return type to new `outgoing-payment-with-spent-amounts` which extends existing `outgoing-payment` with new required fields `grantSpentDebitAmount` and `grantSpentReceiveAmount`.

## Context
fixes: https://github.com/interledger/open-payments/issues/469

progress towards: https://github.com/interledger/rafiki/issues/2678

### DIFF
--- a/.changeset/shy-needles-brush.md
+++ b/.changeset/shy-needles-brush.md
@@ -1,5 +1,5 @@
 ---
-'@interledger/open-payments': major
+'@interledger/open-payments': minor
 ---
 
 Changes POST /outgoing-payment return to new 'outgoing-payment-with-spent-amounts' type with grantSpentDebitAmount and grantSpentReceiveAmount
--- a/.changeset/shy-needles-brush.md
+++ b/.changeset/shy-needles-brush.md
@@ -1,0 +1,5 @@
+---
+'@interledger/open-payments': major
+---
+
+Changes POST /outgoing-payment return to new 'outgoing-payment-with-spent-amounts' type with grantSpentDebitAmount and grantSpentReceiveAmount
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -967,7 +967,6 @@ components:
           metadata:
             description: Thank you for your purchase at ShoeShop!
             externalRef: INV2022-8943756
-      additionalProperties: false
       properties:
         id:
           type: string
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -967,6 +967,7 @@ components:
           metadata:
             description: Thank you for your purchase at ShoeShop!
             externalRef: INV2022-8943756
+      additionalProperties: false
       properties:
         id:
           type: string
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -1112,7 +1112,7 @@ components:
           description: The total amount that has been sent under this outgoing payment.
         grantSpentDebitAmount:
           $ref: './schemas.yaml#/components/schemas/amount'
-          description: The amount already deducted from the sender's account against the outgoing payment.
+          description: The total amount already deducted from the sender's account using the current outgoing payment grant.
         grantSpentReceiveAmount:
           $ref: './schemas.yaml#/components/schemas/amount'
           description: The total amount already received against the outgoing payment.
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -1045,6 +1045,27 @@ components:
             value: '2600'
             assetCode: USD
             assetScale: 2
+          createdAt: '2022-03-12T23:20:50.52Z'
+          updatedAt: '2022-04-01T10:24:36.11Z'
+          metadata:
+            description: APlusVideo subscription
+            externalRef: 'customer: 847458475'
+        - id: 'https://ilp.rafiki.money/outgoing-payments/8c68d3cc-0a0f-4216-98b4-4fa44a6c88cf'
+          walletAddress: 'https://ilp.rafiki.money/alice/'
+          failed: false
+          receiver: 'https://ilp.rafiki.money/aplusvideo/incoming-payments/45d495ad-b763-4882-88d7-aa14d261686e'
+          receiveAmount:
+            value: '2500'
+            assetCode: USD
+            assetScale: 2
+          debitAmount:
+            value: '2600'
+            assetCode: USD
+            assetScale: 2
+          sentAmount:
+            value: '2500'
+            assetCode: USD
+            assetScale: 2
           grantSpentReceiveAmount:
             value: '2500'
             assetCode: USD
@@ -1088,10 +1109,10 @@ components:
           description: The total amount that has been sent under this outgoing payment.
         grantSpentDebitAmount:
           $ref: './schemas.yaml#/components/schemas/amount'
-          description: The total amount already deducted from the sender's account using the current outgoing payment grant.
+          description: The total amount successfully deducted from the sender's account using the current outgoing payment grant.
         grantSpentReceiveAmount:
           $ref: './schemas.yaml#/components/schemas/amount'
-          description: The total amount already received (by all receivers) against the current outgoing payment grant.
+          description: The total amount successfully received (by all receivers) using the current outgoing payment grant.
         metadata:
           type: object
           description: Additional metadata associated with the outgoing payment. (Optional)
@@ -1112,8 +1133,6 @@ components:
         - sentAmount
         - createdAt
         - updatedAt
-        - grantSpentDebitAmount
-        - grantSpentReceiveAmount
     quote:
       title: Quote
       description: A **quote** resource represents the quoted amount details with which an Outgoing Payment may be created.
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -1090,7 +1090,7 @@ components:
           description: The total amount already deducted from the sender's account using the current outgoing payment grant.
         grantSpentReceiveAmount:
           $ref: './schemas.yaml#/components/schemas/amount'
-          description: The total amount already received against the outgoing payment.
+          description: The total amount already received (by all receivers) against the current outgoing payment grant.
         metadata:
           type: object
           description: Additional metadata associated with the outgoing payment. (Optional)
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -1022,19 +1022,122 @@ components:
     outgoing-payment-with-spent-amounts:
       title: Outgoing Payment With Grant Spent Amounts
       description: 'An **outgoing payment** resource represents a payment that will be, is currently being, or has previously been, sent from the wallet address.'
-      allOf:
-        - $ref: '#/components/schemas/outgoing-payment'
-        - type: object
-          properties:
-            grantSpentDebitAmount:
-              $ref: './schemas.yaml#/components/schemas/amount'
-              description: The amount already deducted from the sender's account against the outgoing payment.
-            grantSpentReceiveAmount:
-              $ref: './schemas.yaml#/components/schemas/amount'
-              description: The total amount already received against the outgoing payment.
-          required:
-            - grantSpentDebitAmount
-            - grantSpentReceiveAmount
+      type: object
+      examples:
+        - id: 'https://ilp.rafiki.money/outgoing-payments/8c68d3cc-0a0f-4216-98b4-4fa44a6c88cf'
+          walletAddress: 'https://ilp.rafiki.money/alice/'
+          failed: false
+          receiver: 'https://ilp.rafiki.money/aplusvideo/incoming-payments/45d495ad-b763-4882-88d7-aa14d261686e'
+          receiveAmount:
+            value: '2500'
+            assetCode: USD
+            assetScale: 2
+          debitAmount:
+            value: '2600'
+            assetCode: USD
+            assetScale: 2
+          sentAmount:
+            value: '2500'
+            assetCode: USD
+            assetScale: 2
+          grantSpentDebitAmount:
+            value: '2600'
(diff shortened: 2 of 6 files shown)


.changeset/shy-needles-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@interledger/open-payments': major
```
BlairCurrey, line 2: breaking change because the rafiki `POST /outgoing-payment` validation will now fail because it is expecting `grantSpentDebitAmount` and `grantSpentReceiveAmount` fields which are not returned.

openapi/resource-server.yaml
```diff
@@ -253,7 +253,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/outgoing-payment'
+                $ref: '#/components/schemas/outgoing-payment-with-spent-amounts'
```
BlairCurrey, line 256: Note that this type is only used for the `POST`. I believe this is what we want based on the task description and since we are already getting these amounts in the create, but I figured I should confirm this direction.

I can imagine it would be unnecessarily expensive to do it in the `GET /outgoing-payments/{id}` and especially `GET /outgoing-payments` and presumably the create is sufficient for whatever is consuming this information?

.changeset/shy-needles-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+'@interledger/open-payments': minor
```
BlairCurrey, line 2: I waivered back and forth on minor/major. I don't think this package has breaking changes exactly. Once we update rafiki to this version the api implementation will be out of date but wont be broken. However, since our validator package uses the spec and checks for the new return type, I expect _that_ to break.

openapi/resource-server.yaml
```diff
@@ -1012,6 +1020,22 @@ components:
...
+          properties:
+            grantSpentDebitAmount:
+              $ref: './schemas.yaml#/components/schemas/amount'
+              description: The amount already deducted from the sender's account against the outgoing payment.
+            grantSpentReceiveAmount:
+              $ref: './schemas.yaml#/components/schemas/amount'
+              description: The total amount already received against the outgoing payment.
+          required:
+            - grantSpentDebitAmount
+            - grantSpentReceiveAmount
```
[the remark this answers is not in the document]
BlairCurrey: Seeing an error in CI: 

> Error:  259:25  error  oas3-valid-media-example  Property "grantSpentDebitAmount" is not expected to be here  paths./outgoing-payments.post.responses[201].content.application/json.examples['New Fixed Send Outgoing Payment for $25'].value

If I comment them out, then it actually tells me it expects these properties. It's like the validator expects the `POST /outgoing-payments` example to fulfill each of the `allOf`. That seems reasonable except it doesnt appear to work that way for our `incoming-payment-with-methods` component.

After investigating further I found that this is because `incoming-payment` allows `additionalProperties` (it's not set and defaults to true).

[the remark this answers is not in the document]
BlairCurrey: Ended up in-lining all the outgoing payment properties https://github.com/interledger/open-payments/pull/474/commits/f74588257d004f9f682a62f2e75941e6fe9f3c2e

openapi/resource-server.yaml
```diff
@@ -1012,6 +1019,125 @@ components:
...
+          description: The total amount that should be received by the receiver when this outgoing payment has been paid.
+        debitAmount:
+          $ref: ./schemas.yaml#/components/schemas/amount
+          description: The total amount that should be deducted from the sender's account when this outgoing payment has been paid.
+        sentAmount:
+          $ref: ./schemas.yaml#/components/schemas/amount
+          description: The total amount that has been sent under this outgoing payment.
+        grantSpentDebitAmount:
+          $ref: './schemas.yaml#/components/schemas/amount'
+          description: The amount already deducted from the sender's account against the outgoing payment.
```
mkurapov, line 1115:
```suggestion
          description: The total amount already deducted from the sender's account using the current outgoing payment grant.
```
Or something like, "cumulative debited amount under the current outgoing payment grant".

openapi/resource-server.yaml
```diff
@@ -1012,6 +1019,125 @@ components:
...
+        grantSpentReceiveAmount:
+          $ref: './schemas.yaml#/components/schemas/amount'
+          description: The total amount already received against the outgoing payment.
```
mkurapov, line 1118:
```suggestion
          description: The total amount already received by receivers against the current outgoing payment grant.
```
There's definitely better wording than this, but I just wanted to highlight that it can be multiple receivers (across multiple outgoing payments) under the same outgoing payment grant

.changeset/shy-needles-brush.md
mkurapov: Yeah this versioning stuff is tough. If we really want to follow semver, major makes sense since upgrading this package will cause requests from the client to break (given an older version of Rafiki). But on the other hand, nothing from the _user_/implementor of the client needs to change... 🤷‍♂️ 

Maybe this is fine for now then, in that case

openapi/resource-server.yaml
```diff
@@ -1012,6 +1019,125 @@ components:
         - sentAmount
         - createdAt
         - updatedAt
+    outgoing-payment-with-spent-amounts:
+      title: Outgoing Payment With Grant Spent Amounts
+      description: 'An **outgoing payment** resource represents a payment that will be, is currently being, or has previously been, sent from the wallet address.'
+      type: object
+      examples:
```
mkurapov, line 1026: Do we need both of these examples?

openapi/resource-server.yaml
```diff
@@ -959,7 +967,6 @@ components:
           metadata:
             description: Thank you for your purchase at ShoeShop!
             externalRef: INV2022-8943756
-      additionalProperties: false
```
mkurapov, line 962: should this be deleted? I feel like we had a discussion around `additionalProperties`, but I don't remember it

openapi/resource-server.yaml
BlairCurrey: Hmmm the second option is more succinct, and I understand it, but I feel like the original suggestion is a bit clearer without any context.

.changeset/shy-needles-brush.md
BlairCurrey: Yeah... to rephrase, the `open-payments` package doesn't break with this change. The `open-api` package, if used, breaks with this change. Which just highlights some weirdness of being tightly coupled via the spec. To be clear, I'm not suggesting there is a better way to do that, I just think it makes this weird. I'll keep as minor for now I guess.

openapi/resource-server.yaml
BlairCurrey: I removed the second example.

For context, I followed the 2 examples in the original `outgoing-payment`. But I don't necessarily think we need 2 examples.

openapi/resource-server.yaml
BlairCurrey: That seems like a good thing to emphasize and yeah "received by receivers" sounds a little clunky... this doesn't remove the usage of `received` and `receivers` but maybe sounds a bit better and makes it clear.

> The total amount already received (by all receivers) against the current outgoing payment grant.

At least these kinda redundant forms are more separated? 

Instead of "received" we could maybe say something like "collected" or "gained", "deposited to " etc. but I feel like "received" is consistent with our word-choice elsewhere. I guess we could use "incoming payment" instead of receiver?

> The total amount already received by all incoming payments against the current outgoing payment grant.

I think the first way is clearer IMO.

BlairCurrey: changed to first way for now but feel free to revise

> The total amount already received (by all receivers) against the current outgoing payment grant.

openapi/resource-server.yaml
BlairCurrey: this was confirmed in a discussion about this task

openapi/resource-server.yaml
BlairCurrey: Actually no, this doesn't need to be deleted. I returned it.  It's left over from when I made `outgoing-payment` an `allOf` with the `outgoing-payment` and the new grant spent amounts. But then I switched to inline everything in a new `outgoing-payment-with-spent-amounts` model. That is what you're thinking of and is described in this comment: https://github.com/interledger/open-payments/pull/474#discussion_r1607102838 And this slack thread: https://interledgerfoundation.slack.com/archives/C06KZCKEEKW/p1716390398470419

Basically, using `allOf` to extend models (in this case extending `outgoing-payment` with the grant spent amounts) requires allowing additional properties. I went back and forth on if I wanted to do it this way. It's more convenient for writing models and removes duplication. We do this elsewhere and its in the open api docs as a way to extend models. However, I didn't really intend to signal that additional properties are actually allowed beyond that (in the docs, in validation, or anything else that consumes the spec). Especially given the recent confusion about the `key: any` in the wallet address docs (which is driven by this).

There is a separate argument to be made (and has been made by some at Fynbos IIRC) that we shouldn't disallow `additionalProperties` by default. There may be a good case there, but to this point we've defaulted to not allowing so that's why I'm hesitant to signal they are allowed merely so we can use `allOf` to extend types in the yaml file instead of inlining.

BlairCurrey: I guess the potential counterpoint to my reasoning is that allowing `additionalProperties` isn’t actually a bad thing - just means we have no opinion on additional properties. In which case we can also enjoy using `allOf` to extend `outgoing-payment`.

openapi/resource-server.yaml
```diff
@@ -1012,6 +1020,100 @@ components:
...
+          description: The total amount that should be deducted from the sender's account when this outgoing payment has been paid.
+        sentAmount:
+          $ref: ./schemas.yaml#/components/schemas/amount
+          description: The total amount that has been sent under this outgoing payment.
+        grantSpentDebitAmount:
+          $ref: './schemas.yaml#/components/schemas/amount'
+          description: The total amount already deducted from the sender's account using the current outgoing payment grant.
+        grantSpentReceiveAmount:
+          $ref: './schemas.yaml#/components/schemas/amount'
+          description: The total amount already received (by all receivers) against the current outgoing payment grant.
```
mkurapov, line 1094: Thinking about this, we should use either "against" or "using" in both of the descriptions.
Also, we should add something that says these amounts are "successfully received" or "successfully deducted", to clear up the fact that these amounts do _not_ include this newly created outgoing payment (since we've just created it, and it hasn't actually sent funds yet)

openapi/resource-server.yaml
```diff
@@ -1012,6 +1020,100 @@ components:
...
+            assetCode: USD
+            assetScale: 2
+          grantSpentDebitAmount:
+            value: '2600'
+            assetCode: USD
+            assetScale: 2
+          grantSpentReceiveAmount:
+            value: '2500'
+            assetCode: USD
+            assetScale: 2
```
mkurapov, line 1051: I think these properties should be optional, since in the future, we will only allow either passing in `limits.receiveAmount` or `limits.debitAmount`, but not both: 
- https://github.com/interledger/open-payments/issues/471

Likewise, this would mean that in the example, only one or the other property will be present.

openapi/resource-server.yaml
mkurapov: What is the default behaviour if we don't include `additionalProperties` at all? Does it allow for them? I don't mind leaving it out, as long as it doesn't show `key: any` in the site, since that might be confusing to see on something like the outgoing payment creation API.

We can keep this for now and then at some point just go through the whole spec and remove it, if we choose to do so.

mkurapov (commented):
Main change: I think we should have these two properties optional given that linked issue

BlairCurrey:
@mkurapov 

> What is the default behaviour if we don't include additionalProperties at all? 

The short answer is it defaults to `additionalProperties: true`, and this field is what drives the `key: any` in the docs. So I think my current changes jive with how we've (mostly) been doing things and conforms to your expectations for the docs which is something I considered myself.

This is veering into a larger conversation, but zooming out a bit, I'm starting to think we shouldn't specify `additionalProperties: false` unless we actually have a reason to signal that there shouldn't be any additional properties. Which I don't think we have for any of the existing models. Letting `additionalProperties` default to `true` just means we have no opinion on if implementations return additional stuff. But currently we are saying "there should never be any additional properties returned by a quote," for example. Why? I don't really have a defense for this requirement. I can't anticipate why someone would want to do that, but I also don't have a specific reason for why they shouldn't do it.

openapi/resource-server.yaml
BlairCurrey: > What is the default behaviour if we don't include additionalProperties at all?

The short answer is it defaults to additionalProperties: true, and this field is what drives the key: any in the docs. So I think my current changes jive with how we've (mostly) been doing things and conforms to your expectations for the docs which is something I considered myself.

This is veering into a larger conversation, but zooming out a bit, I'm starting to think we shouldn't specify additionalProperties: false unless we actually have a reason to signal that there shouldn't be any additional properties. Which I don't think we have for any of the existing models. Letting additionalProperties default to true just means we have no opinion on if implementations return additional stuff. But currently we are saying "there should never be any additional properties returned by a quote," for example. Why? I don't really have a defense for this requirement. I can't anticipate why someone would want to do that, but I also don't have a specific reason for why they shouldn't do it. If we are removing some capability I think we need to have a good reason.

openapi/resource-server.yaml
BlairCurrey: made them optional and returned the 2nd examples to show one or the other

openapi/resource-server.yaml
BlairCurrey: Changed to:

> description: The total amount successfully deducted from the sender's account using the current outgoing payment grant.

and:

> description: The total amount successfully received (by all receivers) using the current outgoing payment grant.

PR #484, opened 2024-07-03
chore(open-payments): parse error objects, and return error codes

### LINKED ISSUE #482: Expose `GNAPErrorCode` to `OpenPaymentsClientError` in client
Context: Web Monetization extension

If a user revokes key from wallet, all subsequent requests error with `HTTP 400` + `invalid_client`. But this code isn't available to client to check whether it's a 400 due to `invalid_client` or some other reason. The response body says `invalid_client`, but we can't access the body on error.

Exposing this in `OpenPaymentsClientError` would help us gracefully handle https://github.com/interledger/web-monetization-extension/issues/332

mkurapov:
Now that we are returning the following error response from the AS:
```json
{
    "error": {
        "code": "invalid_client",
        "description": "could not determine client"
    }
}
```

we now populate `code` and `description` fields in the `OpenPaymentsClientError` with the correct values.

## Context
- Fixes #482

### DIFF
--- a/packages/open-payments/src/client/index.test.ts
+++ b/packages/open-payments/src/client/index.test.ts
@@ -75,6 +75,7 @@ describe('Client', (): void => {
     })
 
     test('throws error if could not load private key as Buffer', async (): Promise<void> => {
+      expect.assertions(2)
       try {
         await createAuthenticatedClient({
           logger: silentLogger,
@@ -92,6 +93,7 @@ describe('Client', (): void => {
     })
 
     test('throws error if could not load private key', async (): Promise<void> => {
+      expect.assertions(2)
       try {
         await createAuthenticatedClient({
           logger: silentLogger,
@@ -116,6 +118,7 @@ describe('Client', (): void => {
     `(
       'throws an error if both authenticatedRequestInterceptor and privateKey or keyId are provided',
       async ({ keyId, privateKey }) => {
+        expect.assertions(2)
         try {
           // @ts-expect-error Invalid args
           await createAuthenticatedClient({
--- a/packages/open-payments/src/client/requests.test.ts
+++ b/packages/open-payments/src/client/requests.test.ts
@@ -2,6 +2,7 @@ import {
   createHttpClient,
   deleteRequest,
   get,
+  handleError,
   HttpClient,
   post,
   requestShouldBeAuthorized,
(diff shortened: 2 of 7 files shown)


packages/open-payments/src/client/index.test.ts
```diff
@@ -75,6 +75,7 @@ describe('Client', (): void => {
     })
 
     test('throws error if could not load private key as Buffer', async (): Promise<void> => {
+      expect.assertions(2)
```
mkurapov, line 78: adding some missing assertions

packages/open-payments/src/client/requests.ts
```diff
@@ -158,23 +158,34 @@ const handleError = async (
...
+    let responseBody:
+      | {
+          error: { description: string; code: string }
+        }
+      | string
+      | undefined
 
     try {
-      responseBody = (await error.response.json()) as { message: string }
+      responseBody = await error.response.text()
```
mkurapov, line 174: once we start returning object response from the RS, this can be just `await error.response.json()`, but this handles both

packages/open-payments/src/client/requests.test.ts
```diff
@@ -817,4 +818,151 @@ describe('requests', (): void => {
...
+            errors: [{ message: 'invalid request' }]
+          },
+          requestType: 'POST',
+          url
+        })
+      } catch (error) {
+        assert.ok(error instanceof OpenPaymentsClientError)
+        expect(error.message).toBe('Error making Open Payments POST request')
+        expect(error.description).toBe('Could not validate OpenAPI response')
+        expect(error.code).toBeUndefined()
```
sidvishnoi, line 922: Can this be `validation_failed` or something?

packages/open-payments/src/client/requests.ts
```diff
@@ -158,23 +158,34 @@ const handleError = async (
...
   let validationErrors
+  let errorCode
 
   const { HTTPError } = await import('ky')
 
   if (error instanceof HTTPError) {
-    let responseBody
+    let responseBody:
+      | {
+          error: { description: string; code: string }
```
sidvishnoi, line 168: The `responseBody.error?.code` tests below suggest these could be undefined.
```suggestion
          error: { description?: string; code?: string }
```

sidvishnoi (approved):
LGTM. Thanks @mkurapov!

packages/open-payments/src/client/requests.ts
mkurapov: Down below, I'm primarily checking that `responseBody.error` is defined, and if it is, I'm assuming `error.description` and `error.code` are indeed defined.

I added a [commit](https://github.com/interledger/open-payments/pull/484/commits/7a78eb91eb0c232e21aa349c5a84a5b8c99b798b) to simplify the assertion

PR #480, opened 2024-06-04
chore: export OutgoingPaymentWithSpentAmounts

BlairCurrey:
## Changes proposed in this pull request

exports `OutgoingPaymentWithSpentAmounts`  type

## Context

missed in  https://github.com/interledger/open-payments/pull/474

PR #488, opened 2024-07-08
fix(open-payments): change outgoingPayment.create return type

BlairCurrey:

- updates return type of create outgoing payment method on the open payments client

## Context

fixes issue raised in this comment: https://github.com/interledger/open-payments/pull/480#issuecomment-2202353862

looks like it was probably first encountered here https://github.com/interledger/web-monetization-extension/pull/386/files

### DIFF
--- a/.changeset/eight-cups-reflect.md
+++ b/.changeset/eight-cups-reflect.md
@@ -1,5 +1,5 @@
 ---
-'@interledger/open-payments': major
+'@interledger/open-payments': patch
 ---
 
-Changes return type of outgoingPayment.create to OutgoingPaymentWithSpentAmounts to reflect correct return type. Consumers should update code to account for new return type as needed.
+Changes return type of outgoingPayment.create to OutgoingPaymentWithSpentAmounts to reflect correct return type.
--- a/.changeset/eight-cups-reflect.md
+++ b/.changeset/eight-cups-reflect.md
@@ -1,0 +1,5 @@
+---
+'@interledger/open-payments': major
+---
+
+Changes return type of outgoingPayment.create to OutgoingPaymentWithSpentAmounts to reflect correct return type. Consumers should update code to account for new return type as needed.
--- a/packages/open-payments/src/client/outgoing-payment.ts
+++ b/packages/open-payments/src/client/outgoing-payment.ts
@@ -10,6 +10,7 @@ import {
   getRSPath,
   OutgoingPayment,
   OutgoingPaymentPaginationResult,
+  OutgoingPaymentWithSpentAmounts,
   PaginationArgs
 } from '../types'
 import { get, post } from './requests'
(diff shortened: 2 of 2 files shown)


.changeset/eight-cups-reflect.md
```diff
@@ -0,0 +1,5 @@
+---
+'@interledger/open-payments': major
+---
+
+Changes return type of outgoingPayment.create to OutgoingPaymentWithSpentAmounts to reflect correct return type. Consumers should update code to account for new return type as needed.
```
BlairCurrey, line 5: Major since it may break consumers. such as here perhaps: https://github.com/interledger/web-monetization-extension/pull/386/files

sidvishnoi:
Why do we have two interfaces - `OutgoingPaymentWithSpentAmounts` and `OutgoingPayment`?
The only difference I see between them are two keys: `"grantSpentDebitAmount" | "grantSpentReceiveAmount"`, that are optional. So, I think we can rename `OutgoingPaymentWithSpentAmounts` to `OutgoingPayment` and keep only one of them (and won't be a breaking change either, as optional keys got added).

.changeset/eight-cups-reflect.md
mkurapov: Isn't it technically just adding new optional fields though? i.e. it wouldn't necessarily break current consumers

BlairCurrey:
> Why do we have two interfaces - `OutgoingPaymentWithSpentAmounts` and `OutgoingPayment`? 

Originally we added a different type for the `POST /outgoing-payment` endpoint because we wanted to return these additional fields on that endpoint only. `GET /outgoing-payment` will never have these fields, for example.

I felt like having them as optional properties on the other `/outgoing-payment` endpoints was misleading because we know they will never be defined. Although it would still adhere to the spec, since they are optional, and does simplify the types. 🤔

.changeset/eight-cups-reflect.md
BlairCurrey: could break at build, no?

mkurapov:
I think having an explicit interface is safer (and this still can still be a `minor` version bump _I think_, since its just net new fields). 

We should have the types for the operations adhere close to the spec as possible -> in this case, if we end up making `grantSpentDebitAmount` required (or add a new field, or anything else to `OutgoingPayment` like @BlairCurrey said, it might be misleading/cause issues)

.changeset/eight-cups-reflect.md
sidvishnoi: Just in case, it won't break for extension  (I aliased the import there). So I'm happy with a patch change for this.

BlairCurrey: yeah, you're right. updated to patch

mkurapov (approved):

sidvishnoi (approved):

Issue #495, opened 2024-07-30
Docs: update badge colors in sidenav for api docs


melissahenderson:
The API docs use [badges](https://starlight.astro.build/guides/sidebar/#badges) in the sidenav TOC so you can quickly see the different methods (POST, GET, DELETE). The GET badge is green and the POST badge is blue.

When you open an API doc for a GET or POST call, the color used for the method at the top of the page is the inverse. 

We should flip the colors used in the sidenav so that GET is blue and POST is green.

melissahenderson:
Taken care of by Brad in his PR

Issue #407, opened 2024-01-04
Add glossary to Open Payments docs


melissahenderson:
A glossary file exists in the OP project today (drafts > glossary.mdx). It needs to be completed/expanded, then made public.

PR #496, opened 2024-08-01
Issues 407 and 495

brad-dow:

#407 : Added Glossary - removed from draft status, finished definitions, and put into sidebar
#495 : Fixed badge colors, swapping blue and green


hajjimo (approved):
Two minor things that are not showstoppers:

The badge colors are not the same shade of blue and green as the API docs. We could do custom styling on the badges, but that may be overkill. At least the badges and API docs are better aligned now. 

I'm conflicted about whether we should include Identity Provider in the glossary. It's not really part of Open Payments, and in that sense, it's fine that we don't have an entry for it. At the same time, we do delve into the concept on the Grant negotiation and authorization page, and in that regard, I wonder if it warrants an entry.

docs/src/content/docs/snippets/wallet-get-info.mdx
```diff
@@ -11,7 +11,7 @@ import Js from '/src/partials/js-prerequisites.mdx'
 
 For a client to request a grant from an authorization server, the client must first verify the validity of the wallet address and get the URL of the wallet's authorization server.
 
-These code snippets enable an authenticated client to verify a wallet address, get the basic information required to construct a new transaction, and discover the auth server’s URL. Note that unauthenticated clients can also call the Open Payments API to get wallet address information.
+These code snippets enable an authenticated client to verify a wallet address, get the basic information required to construct a new transaction, and discover the auth server’s URL. Note that unauthenticated clients can also call the Open Payments APIs to get wallet address information.
```
melissahenderson, line 14: In this instance, unauthenticated clients can only call the Wallet Address Server API, not the Resource Server or Auth Server APIs. The plural should be changed back to singular and we need to figure out how to actually refer to these APIs in the docs.

docs/src/content/docs/resources/glossary.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+An account servicing entity provides and maintains a payment account for a payer and payee, and is a regulated entity in the country/countries it operates.
+
+## Authorization server (AS)
+
+An authorization server issues tokens to clients, which can then be used to perform authorized actions on resource servers. In the context of Open Payments, the authorization server grants permission for a client to access the Open Payments APIs and the `incoming-payment`, `quote`, and `outgoing-payment` resources. Open Payments leverages GNAP as the mechanism for delegating authorization.
+
+## Client
+
+A client is an application or service that interacts with the authorization server to obtain grants and tokens. A client uses these tokens to access resources or perform transactions on behalf of a user or system.
```
melissahenderson, line 17: I don't know if this is overkill. Just a suggestion. 

A client is an application or service, such as a mobile or web app, that interacts with an authorization server to obtain grants and tokens. A client uses these tokens to access resources on a resource server to perform actions, such as retrieving transaction history and setting up payments, on behalf of a user or system.

docs/src/content/docs/resources/glossary.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+## Grant Negotiation and Authorization Protocol (GNAP)
+
+The Grant Negotiation Authorization Protocol (GNAP) defines a mechanism for delegating authorization to a piece of software, and conveying the results and artifacts of that delegation to the software. This delegation can include access to a set of APIs as well as subject information passed directly to the software. For more information, see the [specification](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol-12).
+
+## Incoming payment resource
+
+An incoming payment resource is an object that represents a payment being received by an entity. This resource contains information about the incoming payment, such as the amount, currency, receiver's wallet address, and payment status. It is used to track and manage payments that are expected to or have been received.
+
+## Interactive grant
```
melissahenderson, line 27: Building off Mohammed's comment about whether we should include an entry for this in the glossary...

I think a good compromise may be to remove the entry, and in the outgoing payment resource entry, link the word "consent" to /introduction/grants/#identity-providers. 

Within that section, there's a tooltip that defines an interactive grant as requiring explicit interaction/consent from the resource owner before an access token can be issued. This way the reader can see the grant along with the thing that manages consent -- the identity provider.

docs/src/content/docs/resources/glossary.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+## Outgoing payment resource
+
+An outgoing payment resource is an object that represents a payment being sent by an entity. This resource contains information about the outgoing payment, such as the amount, currency, receiver's wallet address, and payment status. Outgoing payment resources require explicit consent from the sender, obtained through an interactive grant, and serves as a mechanism to separate payment instructions from payment execution.
+
+## Quote resource
+
+A quote resource is an object that represents a potential payment being received by an entity. This resource contains information about the potential payment, but is mainly used to indicate the total cost, including any applicable fees, to make the payment. The quote resource also serves as a commitment from the sender's ASE to deliver a particular amount to the receiver's ASE and it only valid for a limited time.
```
melissahenderson, line 41: Link to /introduction/op-concepts/#quote

The resource is created by the sender's ASE, on their resource server, after the incoming payment resource is created by the recipient's ASE.

melissahenderson:
@hajjimo  I see what you mean about the color. If it was something we could easily adjust, I'd say we should change it. But it requires HJ to customize some stuff.

docs/src/content/docs/resources/glossary.mdx
brad-dow: I like that approach. Removing entry for "interactive grant" and linking to identity providers in the "outgoing payment resource" entry.

docs/src/content/docs/resources/glossary.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+## Resource server (RS)
+
+A resource server hosts protected resources and enforces access controls based on the tokens provided by the authorization server. In the context of Open Payments, the resource server manages access to three payment-related resources (`incoming-payment`, `quote`, and `outgoing-payment`), ensuring that only authorized clients can perform actions through the Open Payments APIs.
```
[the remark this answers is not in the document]
brad-dow: I don't have strong feelings either. Because the first sentence is a non-Open Payments-specific definition, that second sentence really sounded like a generic resource server statement before I added that qualifier. 

Authorization Server is structured the same way.

docs/src/content/docs/snippets/wallet-get-info.mdx
brad-dow: I'm going to switch back to singular and refer to it as the "Wallet Address API". That's what it's called in the OpenAPI spec: https://github.com/interledger/open-payments/blob/main/openapi/wallet-address-server.yaml

melissahenderson (approved):

PR #497, opened 2024-08-20
feat: add access-identity to auth-server.yaml

desimmons:

This PR is a proposal for a new grant type: `access-identity`

## Context

As part of the [Open Payments community call](https://docs.google.com/document/d/1b3caT7lWAaTA-rZMI3YTDBgInGWVa7JtMuO17vG84kg/edit?usp=sharing), the need for a user-authentication grant was identified by two participants. One of the actions from the latest call was to create proposals. This is one such proposal.

## TODO

- [ ] Update `resource-server.yaml`
- [ ] Create basic example docs

### DIFF
--- a/openapi/auth-server.yaml
+++ b/openapi/auth-server.yaml
@@ -313,6 +313,7 @@ components:
         - $ref: '#/components/schemas/access-incoming'
         - $ref: '#/components/schemas/access-outgoing'
         - $ref: '#/components/schemas/access-quote'
+        - $ref: '#/components/schemas/access-identity'
       description: The access associated with the access token is described using objects that each contain multiple dimensions of access.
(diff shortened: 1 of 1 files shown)


openapi/auth-server.yaml
```diff
@@ -375,6 +376,27 @@ components:
         - type
         - actions
         - identifier
+    access-identity:
```
raducristianpopa, line 379: Based on the discussion from the Open Payments Community call, this would also require an additional property that will need to be provided - the user wallet address. This way, the wallet provider can validate if the connected user (on their side) is the owner of the wallet address when the user is redirected to the IdP page.

melissahenderson:
Subject-based grant updates care for this issue. Closing.

PR #498, opened 2024-08-23
feat: add `estimatedExchangeRate` property to quote response

### LINKED ISSUE #491: Return `estimatedExchangeRate` for quotes
## Context

Originally from @raducristianpopa particularly in the WM context, there is no good way to determine what is the minimum debitAmount for a quote that will satisfy at least a minimum one unit of value for the receiving wallet address:

> What would be the best way to check what is the minimum amount that can be sent in a cross-currency scenario - debitAmount?
When sending from USD to EUR, in most cases $0.01 is not going to equal 0.01EUR. 
I thought about using receiveAmount but this will not work - the grant will need the limits in the receiver’s asset.
Probe through quoting -  create quotes until one is created successfully? - For WM this will lead to some delays in the flow
Probe through quoting with an exponential increase for the amount?
If the sender’s asset is ZAR and the receiver’s asset is USD we will probably have to make around 16-18 quote requests (depends on the exchange rate as well):
0.01 ZAR = 0.00… USD
0.19 ZAR = 0.01 USD
In the example above, the extension will have to make 18 quote requests until it reaches an amount that can be represented in USD.

What we came up with in the Open Payments catch up call is to expose an `estimatedExchangeRate` on the quote POST & GET response.

## Todos
- [ ] Update the resource server spec to include an optional `estimatedExchangeRate: number` in the quote

raducristianpopa:

- Closes #491

### DIFF
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -1193,6 +1193,10 @@ components:
           description: "The total amount that should be deducted from the sender's account when the corresponding outgoing payment has been paid. "
         method:
           $ref: '#/components/schemas/payment-method'
+        estimatedExchangeRate:
+          type: number 
+          description: 'Estimated probed exchange rate over the path (ILP specific).'
+          readOnly: true
         expiresAt:
           type: string
           description: The date and time when the calculated `debitAmount` is no longer valid.
--- a/openapi/resource-server.yaml
+++ b/openapi/resource-server.yaml
@@ -1194,7 +1194,7 @@ components:
         method:
           $ref: '#/components/schemas/payment-method'
         estimatedExchangeRate:
-          type: number 
+          type: number
           description: 'Estimated probed exchange rate over the path (ILP specific).'
           readOnly: true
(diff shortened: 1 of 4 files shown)


openapi/resource-server.yaml
```diff
@@ -1193,6 +1193,10 @@ components:
           description: "The total amount that should be deducted from the sender's account when the corresponding outgoing payment has been paid. "
         method:
           $ref: '#/components/schemas/payment-method'
+        estimatedExchangeRate:
+          type: number 
+          description: 'Estimated probed exchange rate over the path (ILP specific).'
+          readOnly: true
```
raducristianpopa, line 1199: How will the `estimatedExchangeRate` be returned in this case? IIRC it should take into consideration exchange rate and fees (if they exist). Will this be computed in the RS and returned as a ready to use value for the consumer?

mkurapov:
@raducristianpopa closing in favour of the `minSendAmount` feature

PR #504, opened 2024-10-07
spec: add user and subject information to grant request/response

sabineschaller:

As discussed during the Open Payments call, I added the user and subject objects to the grant request and just the subject object to the grant response.

I'm not sure that subject is actually required, because in a first step, we don't want to request subject information but rather just identify the user. But we can discuss on here or on a call.

### DIFF
--- a/openapi/auth-server.yaml
+++ b/openapi/auth-server.yaml
@@ -44,6 +44,27 @@ paths:
                     required:
                       - access_token
                       - continue
+                  - properties:
+                      subject:
+                        type: object
+                        description: Subject (user) information as requested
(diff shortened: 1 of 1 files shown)


openapi/auth-server.yaml
```diff
@@ -134,8 +187,22 @@ paths:
                           - read
                         identifier: 'http://ilp.rafiki.money/bob'
                   client: 'https://webmonize.com/.well-known/pay'
+              Indentifying the user:
```
desimmons, line 190: Nit: misspelling of "Identifying"

PR #517, opened 2024-11-13
chore: routine dependency bumps

JoblersTune:

Updating dependencies for the docs.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -10,16 +10,16 @@
     "astro": "astro"
   },
   "dependencies": {
-    "@astrojs/starlight": "^0.24.5",
-    "@interledger/docs-design-system": "^0.5.0",
-    "astro": "4.11.3",
-    "mermaid": "^10.9.1",
-    "sharp": "^0.33.4",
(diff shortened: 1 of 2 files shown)


melissahenderson:
Does the astro config file also need the following?

```
      plugins: [
        starlightLinksValidator({
          errorOnLocalLinks: false,
        }),
      ],
```

This was in the Rafiki update, which is why I ask.

JoblersTune:
> Does the astro config file also need the following?
> 
> ```
>       plugins: [
>         starlightLinksValidator({
>           errorOnLocalLinks: false,
>         }),
>       ],
> ```
> 
> This was in the Rafiki update, which is why I ask.

@melissahenderson this is only needed in docs that contain localhost addresses. So for Rafiki because we have the local playground docs we have links that are localhost links which causes the links validator to fail. I think that's the only docs project that uses localhost links. So no, the links validator is passing here.

The introduction of this feature was because usually localhost links are for dev purposes so including them in a production setup is usually an accident.

melissahenderson (approved):

PR #519, opened 2024-11-13
feat: add gnap error response schema to spec

### LINKED ISSUE #518: Add GNAP error schema to auth server error responses
The GNAP spec outlines its own [format for error responses](https://datatracker.ietf.org/doc/html/rfc9635#name-error-response) from an authorization server. This error format is implemented in the Rafiki authorization server, but the authorization server spec should be updated to specify this error format as well.

It might be useful to also surface these potential responses in our own documentation, instead of nesting it inside of the GNAP specification.

njlie:
- Adds a `gnap-error` schema to the spec, and references them in the error responses of the auth server endpoint specs.

## Context

Closes #518.

### DIFF
--- a/openapi/auth-server.yaml
+++ b/openapi/auth-server.yaml
@@ -74,10 +74,22 @@ paths:
                       uri: 'https://auth.rafiki.money/continue/4CF492MLVMSW9MKMXKHQ'
         '400':
           description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
         '401':
           description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
         '500':
           description: Internal Server Error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
       requestBody:
         content:
           application/json:
@@ -198,10 +210,22 @@ paths:
                       wait: 30
         '400':
           description: Bad Request
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
         '401':
           description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
         '404':
           description: Not Found
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/gnap-error'
       requestBody:
         content:
           application/json:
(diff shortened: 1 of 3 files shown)


openapi/auth-server.yaml
```diff
@@ -532,6 +588,26 @@ components:
...
+              enum:
+                - invalid_request
+                - invalid_client
+                - invalid_interaction
+                - invalid_rotation
+                - invalid_continuation
+                - user_denied
+                - request_denied
+                - unknown_interaction
+                  - too_fast
```
oana-lolea, line 610: I think `too_fast` shouldn't be indented

openapi/auth-server.yaml
```diff
@@ -532,6 +588,26 @@ components:
             - debitAmount
         - required:
             - receiveAmount
+    gnap-error:
```
mkurapov, line 591: what do you think of having separate objects per error, and then using specific ones per-response?
e.g. we wouldn't have  `too_fast` in a normal grant request, just for grant continuation.

Maybe that is too specific, but curious to hear your opinion

njlie: I considered that, but it felt at times overly granular to have a schema for an error that was only referenced once - `too_fast` is a pretty good example of that, as it's only thrown by the grant continuation endpoint.

In hindsight though, I think it would be more useful to use the specific ones per-response. Biggest reason to me being that these routes may throw the same code for different reasons, primarily 401 could be thrown by the signature middleware or the actual business logic...

oana-lolea (approved):

oana-lolea (approved):

mkurapov:
@njlie given we verify the Open Payments API responses in the Open Payments client, this actually might be breaking if we are returning incorrect errors in the Rafiki AS, particularly if an integrator is on an old version of Rafiki. Is that correct?

mkurapov:
I was just taking look at the open PRs, and I think we just need to update the client here @njlie

njlie:
> I was just taking look at the open PRs, and I think we just need to update the client here @njlie

Does it need to be updated? I think the regenerated types should be enough for the client to start accepting the new error responses as valid.

> @njlie given we verify the Open Payments API responses in the Open Payments client, this actually might be breaking if we are returning incorrect errors in the Rafiki AS, particularly if an integrator is on an old version of Rafiki. Is that correct?

Old comment, but yes it's breaking in the sense that an outdated client won't be able to accept the new responses, but at least it doesn't break the happy path.

mkurapov (approved):

PR #524, opened 2024-12-04
docs: mi/170/add-use-cases

hajjimo:
Added use case section to overview page.  Will link to remaining guides as they become available.

### DIFF
--- a/docs/src/content/docs/introduction/overview.mdx
+++ b/docs/src/content/docs/introduction/overview.mdx
@@ -67,3 +67,23 @@ When an ASE adopts the Open Payments standard, clients (applications and other p
 - ASEs and clients to create and deploy custom integrations to communicate with one another
 
 Open Payments aims to simplify and democratize payments by providing a standardized, easy-to-integrate solution. This reduces development effort and enhances financial inclusion by making payment solutions more affordable and accessible to all.
+
+## Use cases
+
+Open Payments facilitates various use cases including but not limited to:
+
+### Peer-to-peer payment transfers
+
+As mentioned above, a sender with an account at one Open Payments-enabled ASE can send a payment to a recipient with an account at a different Open Payments-enabled ASE by sending the payment from the sender’s wallet address to the recipient’s wallet address. Thus, Open Payments brings a similar user experience to payments as email brings to person-to-person communications.
+
+### E-commerce platforms
+
+Open Payments can facilitate one-time payments directly between merchants and customers or via third-party marketplaces. A key difference between conventional e-commerce platforms and those enabled by Open Payments is that the payment function won’t be in the control of a sole payment processor or gateway. Refer to the [one-time payment guide](/guides/make-onetime-payment) for more details.
+
+### Recurring payments
+
+In addition to one-time payments that are common on e-commerce marketplaces, Open Payments can also be used to enable subscriptions or payments made at recurring intervals, such as those that are used for Buy Now and Pay Later (BNPL) plans. Refer to the [recurring payment guide](/guides/make-recurring-payments) for more details.
+
+### Web Monetization
+
+Web Monetization is a new and open API standard for website visitors to pay website owners and content creators directly from the browser to the website. As Open Payments is integrated with Web Monetization, website visitors can access content and support creators without going through a paywall. Website owners and content creators can receive payments without depending on any particular payment processor. Refer to <LinkOut href='https://webmonetization.org'>webmonetization.org</LinkOut> for more details.
--- a/docs/src/content/docs/introduction/overview.mdx
+++ b/docs/src/content/docs/introduction/overview.mdx
@@ -94,7 +94,7 @@ Buy Now and Pay Later (BNPL) plans allow customers to pay for high-priced purcha
 
 ### Web Monetization
 
-<LinkOut href='https://webmonetization.org'>Web Monetization</LinkOut> is a new
+The <LinkOut href='https://webmonetization.org'>Web Monetization</LinkOut> is a new
 standard and API that allows website visitors to pay an amount they choose to a
 participating website with little to no interaction using Open Payments as the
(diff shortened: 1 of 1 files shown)


docs/src/content/docs/introduction/overview.mdx
```diff
@@ -67,3 +67,23 @@ When an ASE adopts the Open Payments standard, clients (applications and other p
 - ASEs and clients to create and deploy custom integrations to communicate with one another
 
 Open Payments aims to simplify and democratize payments by providing a standardized, easy-to-integrate solution. This reduces development effort and enhances financial inclusion by making payment solutions more affordable and accessible to all.
+
+## Use cases
+
+Open Payments facilitates various use cases including but not limited to:
+
+### Peer-to-peer payment transfers
```
melissahenderson, line 75: "Payment transfer" might be confusing. Let's use one or the other.

docs/src/content/docs/introduction/overview.mdx
```diff
@@ -67,3 +67,23 @@ When an ASE adopts the Open Payments standard, clients (applications and other p
...
+
+As mentioned above, a sender with an account at one Open Payments-enabled ASE can send a payment to a recipient with an account at a different Open Payments-enabled ASE by sending the payment from the sender’s wallet address to the recipient’s wallet address. Thus, Open Payments brings a similar user experience to payments as email brings to person-to-person communications.
```
melissahenderson, line 77: The primary audiences for the Open Payments docs are ASEs who want to implement the standard and developers who want to build payments into whatever they're developing. I think this use case should focus on peer-to-peer from those perspectives. In fact, I think you can take the "For example" paragraph out of the "Benefits of Open Payments" section and modify it a bit to make it a use case.

docs/src/content/docs/introduction/overview.mdx
```diff
@@ -67,3 +67,23 @@ When an ASE adopts the Open Payments standard, clients (applications and other p
...
+
+### E-commerce platforms
+
+Open Payments can facilitate one-time payments directly between merchants and customers or via third-party marketplaces. A key difference between conventional e-commerce platforms and those enabled by Open Payments is that the payment function won’t be in the control of a sole payment processor or gateway. Refer to the [one-time payment guide](/guides/make-onetime-payment) for more details.
```
melissahenderson, line 81: Is this use case for e-commerce platforms or for one-time payments? Could an e-commerce platform support both one-time and recurring payments?

I suggest working with a dev familiar with Open Payments -- doesn't have to be Max -- and see if together you can come up with a more robust use case. I do like that you mention how OP-enabled platforms don't necessarily need to rely on a single payment processor/gateway.

melissahenderson: Also, side thought, is the one-time payment guide applicable to peer-to-peer?

docs/src/content/docs/introduction/overview.mdx
```diff
@@ -67,3 +67,23 @@ When an ASE adopts the Open Payments standard, clients (applications and other p
...
+
+### Recurring payments
+
+In addition to one-time payments that are common on e-commerce marketplaces, Open Payments can also be used to enable subscriptions or payments made at recurring intervals, such as those that are used for Buy Now and Pay Later (BNPL) plans. Refer to the [recurring payment guide](/guides/make-recurring-payments) for more details.
```
melissahenderson, line 85: Marketplaces could support both one-time and recurring payments, so let's leave mention of marketplaces out of the description. Also, I think we can assume the reader will know what a recurring payment is, so we don't need to say its a payment made at a recurring interval. Ideally this section would describe just one use case for recurring payments in more detail, so just (for example) BNPL, or a monthly subscription/donation/bill payment.

docs/src/content/docs/introduction/overview.mdx
```diff
@@ -67,3 +67,23 @@ When an ASE adopts the Open Payments standard, clients (applications and other p
...
+
+### Web Monetization
+
+Web Monetization is a new and open API standard for website visitors to pay website owners and content creators directly from the browser to the website. As Open Payments is integrated with Web Monetization, website visitors can access content and support creators without going through a paywall. Website owners and content creators can receive payments without depending on any particular payment processor. Refer to <LinkOut href='https://webmonetization.org'>webmonetization.org</LinkOut> for more details.
```
melissahenderson, line 89: Web Monetization doesn't guarantee there won't be paywalls. That's up to the content owner. Radu is probably a good resource to come up with an idea for this section, since he's on the WM team and is familiar with OP.

melissahenderson (changes requested):

docs/src/content/docs/introduction/overview.mdx
hajjimo: Reworked text using an admixture of the use case on the Rafiki overview page and the "For example" section.

docs/src/content/docs/introduction/overview.mdx
hajjimo: Leveraged text from use case on Rafiki docs and added links to one-time, recurring guides and Max's e-commerce example.  Also added one-time payment guide to peer-to-peer

docs/src/content/docs/introduction/overview.mdx
hajjimo: Added a separate use case for BNPL with a link to the recurring payments guide.

docs/src/content/docs/introduction/overview.mdx
hajjimo: Leveraged text from Web Monetization use case from the Rafiki overview page.

melissahenderson (approved):
We chatted in slack about some changes to Peer-to-peer payments. I'm going to go ahead and approve this PR though. I might need to approve it again if P2P is changed.

PR #525, opened 2024-12-06
feat: added basic site tracking using Umami

JoblersTune:

Added an Umami site tracking script so that we can record page visits for the Open Payments docs.

## Context

This gets the basic infrastructure in place and allows us to track page views. There will be more work needed to add granular event tracking based on clicking links.


huijing (approved):
LGTM

PR #534, opened 2025-01-15
chore: replacing .guide with .dev

JoblersTune:

The docs were moved from .guide to .dev so I'm just updating mentions.


JoblersTune:
I don't want to update the CNAME file without understanding all the implications and ensuring I'm not breaking anything. I think all these changes are safe.

PR #549, opened 2025-02-10
chore: dependency bumps and ci updates

huijing:

This PR bumps dependencies for the Open Payments documentation site and updates the Github actions to their latest versions

### DIFF
--- a/.github/workflows/deploy-docs.yaml
+++ b/.github/workflows/deploy-docs.yaml
@@ -40,7 +40,7 @@ jobs:
           pnpm build
 
       - name: Upload artifact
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-docs.yaml
```diff
@@ -53,4 +53,4 @@ jobs:
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v1
+        uses: actions/deploy-pages@v4
```
mkurapov, line 56: Do you think we should move to Netlify completely (like rafiki.dev)?

mkurapov (approved):

PR #554, opened 2025-02-18
docs: mi/537/further-learning

hajjimo:
Added further learning page under Resources.


hajjimo:
The Open Payments SDK might be a better fit for the Coding section under the Getting involved, as a new separate page under Resources, linked from the split-payments tutorial, or a combination of these.

PR #564, opened 2025-03-12
feat: update outgoing payment limits spec

### LINKED ISSUE #471: Update how we set limits during outgoing payment grant creation

njlie:

* Prevents outgoing payment grant requests where `debitAmount` and `receiveAmount` are both included in the `limits`.

## Context

Closes #471.

### DIFF
--- a/packages/open-payments/src/client/grant.ts
+++ b/packages/open-payments/src/client/grant.ts
@@ -11,7 +11,9 @@ import {
   Grant,
   GrantContinuation,
   GrantRequest,
-  GrantContinuationRequest
+  GrantContinuationRequest,
+  OutgoingPaymentLimitWithDebitAmount,
+  OutgoingPaymentLimitWithReceiveAmount
 } from '../types'
 import { post, deleteRequest } from './requests'
 
@@ -62,8 +64,10 @@ export const createGrantRoutes = (deps: GrantRouteDeps): GrantRoutes => {
         (el) => el.type === 'outgoing-payment'
       )
       if (
-        outgoingPaymentAccess?.limits?.debitAmount &&
-        outgoingPaymentAccess.limits?.receiveAmount
+        (outgoingPaymentAccess?.limits as OutgoingPaymentLimitWithDebitAmount)
+          ?.debitAmount &&
+        (outgoingPaymentAccess?.limits as OutgoingPaymentLimitWithReceiveAmount)
+          ?.receiveAmount
       ) {
         throw new OpenPaymentsClientError('Invalid Grant Request', {
           description: 'Invalid Request'
--- a/packages/open-payments/src/client/grant.ts
+++ b/packages/open-payments/src/client/grant.ts
@@ -12,8 +12,8 @@ import {
   GrantContinuation,
   GrantRequest,
   GrantContinuationRequest,
-  OutgoingPaymentLimitWithDebitAmount,
-  OutgoingPaymentLimitWithReceiveAmount
+  AccessOutgoingWithDebitAmount,
+  AccessOutgoingWithReceiveAmount
 } from '../types'
 import { post, deleteRequest } from './requests'
 
@@ -64,13 +64,14 @@ export const createGrantRoutes = (deps: GrantRouteDeps): GrantRoutes => {
         (el) => el.type === 'outgoing-payment'
       )
       if (
-        (outgoingPaymentAccess?.limits as OutgoingPaymentLimitWithDebitAmount)
+        (outgoingPaymentAccess?.limits as AccessOutgoingWithDebitAmount)
           ?.debitAmount &&
-        (outgoingPaymentAccess?.limits as OutgoingPaymentLimitWithReceiveAmount)
+        (outgoingPaymentAccess?.limits as AccessOutgoingWithReceiveAmount)
           ?.receiveAmount
       ) {
         throw new OpenPaymentsClientError('Invalid Grant Request', {
-          description: 'Invalid Request'
+          description:
+            'Only one of "debitAmount" or "receiveAmount" may be specified.'
         })
       }
       return post(
--- a/packages/open-payments/src/client/grant.ts
+++ b/packages/open-payments/src/client/grant.ts
@@ -53,11 +53,20 @@ export const createGrantRoutes = (deps: GrantRouteDeps): GrantRoutes => {
   }
 
   return {
-    request: (
+    request: async (
       { url }: UnauthenticatedResourceRequestArgs,
       args: Omit<GrantRequest, 'client'>
-    ) =>
-      post(
+    ) => {
+      const outgoingPaymentAccess = args.access_token.access.find(
+        (el) => el.type === 'outgoing-payment'
+      )
+      if (
+        outgoingPaymentAccess?.limits?.debitAmount &&
+        outgoingPaymentAccess.limits?.receiveAmount
+      ) {
+        throw new Error('Invalid Grant Request')
+      }
+      return post(
         baseDeps,
         {
           url,
@@ -67,7 +76,8 @@ export const createGrantRoutes = (deps: GrantRouteDeps): GrantRoutes => {
           }
         },
         requestGrantValidator
-      ),
+      )
+    },
     continue: (
       { url, accessToken }: GrantOrTokenRequestArgs,
       args: GrantContinuationRequest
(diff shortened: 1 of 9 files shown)


packages/open-payments/src/openapi/generated/auth-server-types.ts
```diff
@@ -164,14 +164,22 @@ export interface components {
...
+        interval?: components["schemas"]["interval"];
+        /** @description All amounts are maxima, i.e. multiple payments can be created under a grant as long as the total amounts of these payments do not exceed the maximum amount per interval as specified in the grant. */
+        debitAmount: external["schemas.yaml"]["components"]["schemas"]["amount"];
+      }> &
+      Partial<{
+        receiver?: external["schemas.yaml"]["components"]["schemas"]["receiver"];
+        interval?: components["schemas"]["interval"];
+        /** @description All amounts are maxima, i.e. multiple payments can be created under a grant as long as the total amounts of these payments do not exceed the maximum amount per interval as specified in the grant. */
+        receiveAmount: external["schemas.yaml"]["components"]["schemas"]["amount"];
+      }>;
```
mkurapov, line 182: These types would allow the client to take in both `receiveAmount` or `debitAmount` (because of the old openapi-typescript lib). I've updated it [here](https://github.com/interledger/open-payments/pull/567) (and upgraded to TS 5 [here](https://github.com/interledger/open-payments/pull/569)) so we can get those in first)

raducristianpopa (commented):
I think we should add some runtime assertions as well when we are dealing with mutually exclusive types and not fully rely on TS errors.

packages/open-payments/src/client/grant.ts
```diff
@@ -53,11 +53,20 @@ export const createGrantRoutes = (deps: GrantRouteDeps): GrantRoutes => {
...
-      post(
+    ) => {
+      const outgoingPaymentAccess = args.access_token.access.find(
+        (el) => el.type === 'outgoing-payment'
+      )
+      if (
+        outgoingPaymentAccess?.limits?.debitAmount &&
+        outgoingPaymentAccess.limits?.receiveAmount
+      ) {
+        throw new Error('Invalid Grant Request')
```
mkurapov, line 67: Maybe we can throw an `OpenPaymentsClientError` here

packages/open-payments/src/types.ts
```diff
@@ -160,3 +161,12 @@ export const AccessAction: Record<
...
 })
+
+export type OutgoingPaymentLimitWithDebitAmount = Extract<
+  ASComponents['schemas']['access-outgoing']['limits'],
+  { debitAmount: components['schemas']['amount'] }
+>
+export type OutgoingPaymentLimitWithReceiveAmount = Extract<
+  ASComponents['schemas']['access-outgoing']['limits'],
+  { receiveAmount: components['schemas']['amount'] }
+>
```
raducristianpopa, line 172:
```suggestion

export type AccessOutgoingBase = {
  receiver?: ASComponents['schemas']['receiver']
  interval?: ASComponents['schemas']['interval']
}

export type AccessOutgoingWithDebitAmount = AccessOutgoingBase & {
  debitAmount: ASComponents['schemas']['amount']
  receiveAmount?: never
}

export type AccessOutgoingWithReceiveAmount = AccessOutgoingBase & {
  debitAmount?: never
  receiveAmount: ASComponents['schemas']['amount']
}

export type AccessOutgoingLimits =
  | AccessOutgoingBase
  | AccessOutgoingWithDebitAmount
  | AccessOutgoingWithReceiveAmount

export type AccessOutgoing = {
  type: 'outgoing-payment'
  actions: ('create' | 'read' | 'read-all' | 'list' | 'list-all')[]
  identifier: string
  limits?: AccessOutgoingLimits
}

export type GrantRequestAccessItem =
  | ASComponents['schemas']['access-incoming']
  | AccessOutgoing
  | ASComponents['schemas']['access-quote']
```

With the new `GrantRequestAccessItem`, we can now update `GrantRequest` from
https://github.com/interledger/open-payments/blob/97612b49ceb7d36c818526aaed700c2f81ffd057/packages/open-payments/src/types.ts#L96-L100
to:
```typescript
export type GrantRequest = {
  access_token: { access: GrantRequestAccessItem[] }
  client: ASOperations['post-request']['requestBody']['content']['application/json']['client']
  interact?: ASOperations['post-request']['requestBody']['content']['application/json']['interact']
}
```

Now, if a client will provide both amounts, a TypeScript error will be raised - in their IDE or at transpilation time:
```sh
open-payments/packages/open-payments on  nl/471/outgoing-payment-limits-update [!?] via   v20.18.3 pnpm tsc --noEmit
src/scratchpad.ts:26:15 - error TS2322: Type '{ assetScale: number; assetCode: string; value: string; }' is not assignable to type 'undefined'.

26               receiveAmount: {
                 ~~~~~~~~~~~~~


Found 1 error in src/scratchpad.ts:26

```

packages/open-payments/src/client/grant.ts
```diff
@@ -53,11 +56,24 @@ export const createGrantRoutes = (deps: GrantRouteDeps): GrantRoutes => {
...
-      post(
+    ) => {
+      const outgoingPaymentAccess = args.access_token.access.find(
+        (el) => el.type === 'outgoing-payment'
+      )
+      if (
+        (outgoingPaymentAccess?.limits as OutgoingPaymentLimitWithDebitAmount)
+          ?.debitAmount &&
+        (outgoingPaymentAccess?.limits as OutgoingPaymentLimitWithReceiveAmount)
+          ?.receiveAmount
```
raducristianpopa, line 70: _(if we proceed with the new types)_
```suggestion
        (outgoingPaymentAccess?.limits as OutgoingAccessWithDebitAmount)
          ?.debitAmount &&
        (outgoingPaymentAccess?.limits as OutgoingAccessWithReceiveAmount)
          ?.receiveAmount
```

raducristianpopa (commented):
I hoped that the `openapi-typescript` update will fix the XOR issued that we are facing for mutually exclusive types, but it looks like it is not generating proper XOR for `oneOf`. Right now, TypeScript will not complain if a client still provides both amounts.

Since this is a major bump, we can improve the types a bit. Let me know what you think about this.

packages/open-payments/src/client/grant.ts
njlie: This looks great! Thanks for writing it up. The discussion on native XOR implementation is _very_ spirited...

packages/open-payments/src/types.ts
njlie: This looks great! Thanks for writing it up. The discussion on native XOR implementation is very spirited...

raducristianpopa (commented):
@njlie it looks like `oneOf` is working a bit differently from what I expected when reading the docs - https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/#difference-between-anyof-and-oneof (???)

I think that the issue is with the `interval` and `receiver` properties since they "match" with multiple schemas when using `oneOf`. Reverting to `anyOf` should be the go to solution in this case. Sorry for putting you on the wrong track! 🙇‍♂️

njlie:
> @njlie it looks like `oneOf` is working a bit differently from what I expected when reading the docs - [swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not#difference-between-anyof-and-oneof](https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/#difference-between-anyof-and-oneof) (???)
> 
> I think that the issue is with the `interval` and `receiver` properties since they "match" with multiple schemas when using `oneOf`. Reverting to `anyOf` should be the go to solution in this case. Sorry for putting you on the wrong track! 🙇‍♂️

I also interpreted it the same way as you, at first - so I thought I could "debug" what the issue with the validation was. But this explains why it's been so confusing. Thanks for flagging this anyways 👍

raducristianpopa (approved):

mkurapov:
Interesting, so `oneOf` to properly work needs all the keys in the objects to be exclusive from one another? Pretty odd but good to know 🤷

PR #584, opened 2025-04-07
feat: add error object responses

### LINKED ISSUE #573: Add error object responses for resource server
### Overview

We can improve the errors response types for the resource server spec.
Instead of specifying just the error code that could be thrown under a particular route, we can also add an error object to expect:

```
{ 
    "error": {
         "code": ... (string),
         "description": ... (string),
         "details": { ... } (object)
    }
}
```

The `details` field can be optional in the error. 

Something similar was done for the authorization server was done in #518. For the RS however, we can just return a generic error object, instead of listing out all of the different errors that can be thrown for now.

### Todo
- [ ] Add error object to error responses in the resource server spec
- [ ] Handle & add `details?: Record<string, unknown>` as a property of the `OpenPaymentsClientError` in the client package

cozminu:

 - add error object responses
 - add Swagger UI for preview when editing

## Context

fixes #573


openapi/resource-server.yaml
```diff
@@ -1253,6 +1289,26 @@ components:
...
       description: Forbidden
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/error-response'
+          examples:
+            Error response with extra details:
+              $ref: '#/components/examples/error-response-full'
+            Error response with no details:
+              $ref: '#/components/examples/error-response-minimal'
```
mkurapov, line 1300:
```suggestion
            Standard error response:
              $ref: '#/components/examples/error-response-full'
            Detailed error response:
              $ref: '#/components/examples/error-response-minimal'
```

openapi/resource-server.yaml
```diff
@@ -1234,6 +1238,38 @@ components:
...
           ilpAddress: string
           sharedSecret: string
+    error-response:
+      type: object
+      properties:
+        error:
+          type: object
+          properties:
+            code:
+              type: number
```
mkurapov, line 1248: Like in the AS spec, code is a string, e.g. `invalid_request `. In the future, we can have a standardized error codes for the RS, but for now, making this a string is sufficient.

openapi/resource-server.yaml
```diff
@@ -1253,6 +1289,26 @@ components:
...
+        application/json:
+          schema:
+            $ref: '#/components/schemas/error-response'
+          examples:
+            Error response with extra details:
+              $ref: '#/components/examples/error-response-full'
+            Error response with no details:
+              $ref: '#/components/examples/error-response-minimal'
+    '404':
+      description: Not Found
```
mkurapov, line 1302: We should also return objects for the 401, and 400 errors

packages/open-payments/src/openapi/generated/resource-server-types.ts
```diff
@@ -381,6 +381,16 @@ export interface components {
             /** @description The base64 url-encoded shared secret to use when establishing a STREAM connection. */
             sharedSecret: string;
         };
+        "error-response": {
+            error: {
+                code: number;
+                description: string;
+                /** @description Additional details about the error. */
+                details?: {
+                    [key: string]: unknown;
```
mkurapov, line 390: We should expose `details` in the `OpenPaymentsClientError`

mkurapov: Since we are updating the client, we should also do a patch changeset

README.md
```diff
@@ -85,6 +85,26 @@ pnpm i
...
+
+- [Docker](https://docs.docker.com/get-docker/) to run containerized services
+
+#### Installation steps
+
+Start Preview in Swagger UI
+
+```sh
+pnpm viewer:compose up
+```
```
mkurapov, line 98: I like the idea - I'm wondering though if instead, we show how to generate the docs (and the corresponding specs) locally `pnpm --filter openpayments-docs start`. Are you missing something in the generated docs which exists in Swagger?

cozminu: Neither hot-reload, nor refresh doesn't work when the spec is changed. A restart is needed to see the changes. I searched for some way to hot-reload it when the spec changes but no luck. 
I'm changing this Readme section to guide the contributors to run the docs locally.

README.md
```diff
@@ -85,6 +85,26 @@ pnpm i
...
+
+After starting the serive, SwaggerUI will be available at http://localhost:8000
```
mkurapov, line 100:
```suggestion
After starting the service, SwaggerUI will be available at http://localhost:8000
```

openapi/resource-server.yaml
cozminu: done

openapi/resource-server.yaml
cozminu: done

packages/open-payments/src/openapi/generated/resource-server-types.ts
cozminu: done

packages/open-payments/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@interledger/open-payments",
-  "version": "6.14.1",
+  "version": "6.14.2",
```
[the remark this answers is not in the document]
cozminu: reversed, added changeset

viewer/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3.8'
```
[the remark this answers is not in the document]
cozminu: removed

packages/open-payments/src/client/error.ts
```diff
@@ -18,5 +24,6 @@ export class OpenPaymentsClientError extends Error {
     this.status = args.status
     this.code = args.code
     this.validationErrors = args.validationErrors
+    this.details = args.details
```
mkurapov, line 27: We need to pass through `details` during the `if (error instanceof HTTPError)` check in the `handleError` function (`open-payments/src/client/requests.ts`)

.changeset/chatty-turkeys-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+"@interledger/open-payments": patch
```
mkurapov, line 2: we can make this minor, since we are adding new functionality technically

packages/open-payments/src/client/error.ts
cozminu: I was not aware of this one. Added details to the response and also a test

mkurapov:
Just need to run `pnpm format`

mkurapov (approved):

PR #590, opened 2025-04-13
feat(spec): subject to as

### LINKED ISSUE #561: Add `subject` field to AS spec
\*\*Note: we can reuse the changes/commits from \*\*[**https://github.com/interledger/open-payments/pull/504**](https://github.com/interledger/open-payments/pull/504)

During a grant request, a client should be able to specify any subjects they might want to acquire information on from the RS. GNAP achieves this through the subject field, though this is currently unimplemented in the OP spec.

[Subject Information field specification](https://www.rfc-editor.org/rfc/rfc9635.html#name-requesting-subject-informati)

TODOs:

- [ ] Add subject field (with `sub_id_formats` and `sub_ids` keys) to the Grant Request body spec:

Example:

```
"subject":  {
     "sub_id_formats": ["uri"],
     "sub_ids":["<wallet address>"]
}
```

- [ ] Allow `access_token` to be optional in request body
  - [ ] Grant request body should allow for either `access_token` or `subject` to be optional: i.e. a grant request can contain just `access_token`, just `subject`, both `access_token` and `subject`, but cannot be empty

golobitch:
Add subject field to AS spec and generate types

## Context
fixes #561


openapi/auth-server.yaml
```diff
@@ -99,20 +99,28 @@ paths:
...
-                    access:
-                      $ref: '#/components/schemas/access'
                 client:
                   $ref: '#/components/schemas/client'
                 interact:
                   $ref: '#/components/schemas/interact-request'
               required:
-                - access_token
                 - client
+              oneOf:
```
mkurapov, line 108: We can support both `subject` and `access_token` within the same request, so instead of oneOf, we can use `minProperties: 1` maybe? Docs: https://swagger.io/docs/specification/v3_0/data-models/data-types/#number-of-properties

cozminu: Used anyOf & required to achieve the desired behavior 

https://github.com/interledger/open-payments/blob/cda6313572434d7b6234a04146a17b335c4e2523/openapi/auth-server.yaml#L588-L590

packages/open-payments/src/types.ts
```diff
@@ -82,12 +82,13 @@ export type CreateQuoteArgs =
 export const getASPath = <P extends keyof ASPaths>(path: P): string =>
   path as string
 export type NonInteractiveGrantRequest = {
-  access_token: ASOperations['post-request']['requestBody']['content']['application/json']['access_token']
+  access_token: { access: GrantRequestAccessItem[] }
```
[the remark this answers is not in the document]
cozminu: done, thanks for the heads-up

packages/open-payments/src/types.ts
```diff
@@ -82,20 +82,22 @@ export type CreateQuoteArgs =
 export const getASPath = <P extends keyof ASPaths>(path: P): string =>
   path as string
 export type NonInteractiveGrantRequest = {
```
mkurapov: I think this type is unused

packages/open-payments/src/types.ts
```diff
@@ -81,21 +81,20 @@ export type CreateQuoteArgs =
 
 export const getASPath = <P extends keyof ASPaths>(path: P): string =>
   path as string
-export type NonInteractiveGrantRequest = {
-  access_token: ASOperations['post-request']['requestBody']['content']['application/json']['access_token']
-  client: ASOperations['post-request']['requestBody']['content']['application/json']['client']
-}
+
 export type Grant = {
   access_token: ASComponents['schemas']['access_token']
```
mkurapov:
```suggestion
  access_token?: ASComponents['schemas']['access_token']
```

mkurapov: I think we should also add a method similar to `isFinalizedGrant` (something like `isFinalizedGrantWithSubject`) that checks for the existence of the `subject` field

cozminu: done

mkurapov (approved):

Issue #609, opened 2025-05-29
[DOCS] Improve existing Open Payments Guides


melissahenderson:
Each sub-issue contains the same checklist. Some items may already be done for a particular guide.

melissahenderson:
Taken care of elsewhere

PR #630, opened 2025-07-03
Mi/608/asset code scale

hajjimo:
Addresses issue # 608 

- Added description and example to simplify understanding of assetCode and assetScale 

Docs files:
guides/make-onetime-payment
guides/make-recurring-payments

API:
openapi/schemas.yaml
(The readability of the description remains an issue when expanding the object definitions in the specs.)


openapi/schemas.yaml
```diff
@@ -30,13 +30,13 @@ components:
...
       type: string
-      description: The assetCode is a code that indicates the underlying asset. This SHOULD be an ISO4217 currency code.
+      description: The assetCode is a code that indicates the underlying asset. This SHOULD be an ISO4217 currency code. As an example, the euro is represented as EUR.
     assetScale:
       title: Asset scale
       type: integer
       minimum: 0
       maximum: 255
-      description: The scale of amounts denoted in the corresponding asset code.
+      description: The number of decimal places used to represent the minimum possible value of the asset. As an example, EUR has an asset scale of 2 with a minimum possible value of €0.01.
```
melissahenderson: I think we should also include an example here that more closely matches what you might see in the doc. Maybe something like this tacked on to the end: 

An amount with a value of 2500 and an asset scale of 2 translates to €25.00.

docs/src/content/docs/guides/make-onetime-payment.mdx
```diff
@@ -73,8 +73,8 @@ Create an `incomingPayment` on the recipient wallet's resource server using the
 Add the `incomingAmount` object and define the following properties:
 
 - `value` : the maximum allowable amount that will be paid to the recipient's wallet address.
-- `assetCode` : the ISO 4217 currency code representing the underlying asset used to make the payment.
-- `assetScale` : the scale of amounts denoted in the corresponding asset code.
+- `assetCode` : the ISO 4217 currency code representing the underlying asset used to make the payment. As an example, the euro is represented as EUR.
```
melissahenderson: The schema says this **should** be an ISO 4217 code. Here, it doesn't sound optional. I don't know which one is right. If Open Payments can, or will one day, support cryptocurrencies, then there wouldn't be an ISO 4217 code.

docs/src/content/docs/guides/make-onetime-payment.mdx
```diff
@@ -73,8 +73,8 @@ Create an `incomingPayment` on the recipient wallet's resource server using the
...
+- `assetScale` : the number of decimal places used to represent the minimum value in the corresponding asset code. EUR has an asset scale of 2 with a minimum value of €0.01.
```
melissahenderson: A few thoughts about assetScale and code:

We might have more guides that end up requiring these properties. Instead of defining them in each guide, maybe they're defined in the glossary and linked to within the guides. Then the guides can be changed so that they're more specific. In the one-time payment guide, the snippet uses a value of 1000, code of USD, and scale of 2. But the description of assetCode in the list uses an example of EUR.

hajjimo: We don't explain these properties in the two newer guides, though assetCode and assetScale appear in their example code.  For consistency, we should add descriptions to the newer guides or remove them from the older ones. I'm fine with either direction. If we embed them into the docs, we could include them as glossary entries or create a partial. Either approach seems fine.  Good point about the EUR-USD mismatch.  I was aiming for currency diversity, but we should keep things in sync with the code samples.

docs/src/content/docs/guides/make-onetime-payment.mdx
hajjimo: Technically, since assetCode is defined as a string, it could support non ISO codes.  I'd be curious to know if Rafiki's OP implementation checks if the asset code conforms to 4217.

melissahenderson: https://rafiki.dev/overview/concepts/accounting/#assets

"Should be used whenever possible".

So, ISO 4217 is a best practice and at least from the doc's point of view, Rafiki doesn't check.

docs/src/content/docs/guides/make-onetime-payment.mdx
melissahenderson: Now that we're slowly getting more guides, we definitely need to figure out how to standardize them. That's part of the reason for #609, but there will still be more to do. 

Here's my suggestion for now. Leave the descriptions in the one-time and recurring guides. Make sure the examples match whatever is in the example code. Then also add the terms to the glossary. I think the definitions can match what ends up in the schemas.yml. You can also use the definition in the glossary to link out to the ISO standard.

BlairCurrey:
> Also, i'm not sure what the build error ["Fail if missing new open-payments type"](https://github.com/interledger/open-payments/actions/runs/16050115514/job/45290562580?pr=630) means, but it might be fixed by doing a pnpm install.

Looks like this error is because it generated the types and saw a difference between what was generated and what the current version is. In theory it can be resolved by doing `pnpm --filter open-payments generate:types` and committing. 

Although when I do this locally Im surprised by all the differences - I see this PR just updates the description but there are a lot more changes than that. I'm not sure why. Maybe its just lagging - somehow previous PR changed things but new type generation wasnt enforced. In which case it's unexpected but probably correct. Or the tool that generates it changed. - last update was 4 months ago but not sure if there have been changes since.

melissahenderson (approved):
I'm approving this PR, but I'd like you to get a dev review just to be sure the updates are good for the API-related files.

docs/src/content/docs/guides/make-onetime-payment.mdx
```diff
@@ -73,8 +73,8 @@ Create an `incomingPayment` on the recipient wallet's resource server using the
 Add the `incomingAmount` object and define the following properties:
 
 - `value` : the maximum allowable amount that will be paid to the recipient's wallet address.
-- `assetCode` : the ISO 4217 currency code representing the underlying asset used to make the payment.
-- `assetScale` : the scale of amounts denoted in the corresponding asset code.
+- `assetCode` : the code representing the underlying asset used to make the payment. An ISO 4217 currency code should be used wheverever possible. The ISO 4217 currency code for the US Dollar is represented as USD.
+- `assetScale` : the number of decimal places used to represent the minimum value in the corresponding asset code. USD has an asset scale of 2 with a minimum value of $0.01. An integer amount of 1000 with an `assetCode` of USD and `assetScale` of 2 translates to $10.00
```
raducristianpopa: Thoughts on:


```suggestion
- `assetScale` : the number of decimal places that defines the scale of the smallest divisible unit for the given asset code. It determines how an integer amount is scaled to derive the actual monetary value. For example, USD has an asset scale of 2 with the smallest unit being 0.01. An integer amount of `1000` with an `assetCode` of USD and `assetScale` of 2 translates to $10.00.
```

docs/src/content/docs/resources/glossary.mdx
```diff
@@ -8,6 +8,14 @@ import { LinkOut } from '@interledger/docs-design-system'
 
 An account servicing entity provides and maintains a payment account for a sender and recipient, and is a regulated entity in the country/countries it operates.
 
+## Asset code
+
+An `assetCode` is a code representation of the underlying asset used to make a payment. The `assetCode` should be an <LinkOut href='https://en.wikipedia.org/wiki/ISO_4217'>ISO 4217</LinkOut> currency code. The ISO 4217 currency code representing the US Dollar is USD.
```
raducristianpopa: This should be updated as well (based on the discussions I've seen in the pull request), since the asset code might be something outside of ISO4217.

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -76,8 +76,8 @@ Create an `incomingPayment` on the recipient wallet's resource server using the
 Add the `incomingAmount` object and define the following properties:
 
 - `value` : the maximum allowable amount that will be paid to the recipient's wallet address.
-- `assetCode` : the ISO 4217 currency code representing the underlying asset used to make the payment.
-- `assetScale` : the scale of amounts denoted in the corresponding asset code.
+- `assetCode` : the code representing the underlying asset used to make the payment. An ISO 4217 currency code should be used wheverever possible. The ISO 4217 currency code for the US Dollar is represented as USD.
+- `assetScale` : the number of decimal places used to represent the minimum value in the corresponding asset code. USD has an asset scale of 2 with a minimum value of $0.01. An integer amount of 1000 with an `assetCode` of USD and `assetScale` of 2 translates to $10.00
```
raducristianpopa: _Needs update if the asset scale suggestion that I made above is more clearer._

docs/src/content/docs/resources/glossary.mdx
```diff
@@ -8,6 +8,14 @@ import { LinkOut } from '@interledger/docs-design-system'
...
+
+## Asset scale
+
+The `assetScale` is the number decimal places used to represent the minimum value of an asset. As an example, USD has an asset scale of 2 with a minimum value of $0.01. An interger amount of 1000 with an `assetCode` of USD and an `assetScale` of 2 translates to $10.00.
```
raducristianpopa: _Needs update if the asset scale suggestion that I made above is more clearer._

docs/src/content/docs/guides/make-onetime-payment.mdx
raducristianpopa: On the other hand, I can see that we reuse the asset scale explanation in multiple places. Would it be easier to link to glossary directly instead of having 3 instances of the same explanation throughout the docs?

melissahenderson: I'm making updates in a different PR that actually remove the definitions for these terms from the guides specifically. I didn't mark it here because I didn't want to potentially cause merge conflicts later.

docs/src/content/docs/resources/glossary.mdx
```diff
@@ -8,6 +8,14 @@ import { LinkOut } from '@interledger/docs-design-system'
 
 An account servicing entity provides and maintains a payment account for a sender and recipient, and is a regulated entity in the country/countries it operates.
 
+## Asset code
+
+A code representation of the underlying asset used to make a payment. The `assetCode` can be an <LinkOut href='https://en.wikipedia.org/wiki/ISO_4217'>ISO 4217</LinkOut> currency code such as EUR (euro), MXN (Mexican peso) or USD (US dollar). It can also be used to represent cryptocurrencies such as BTC (Bitcoin), ETH (Ethereum) or USDC (USD stablecoin).
```
raducristianpopa: We are okay with using cryptocurrencies here?

raducristianpopa (commented):
Only one question from my side:

docs/src/content/docs/resources/glossary.mdx
hajjimo: I mentioned cryptocurrencies in the glossary entry for asset code based on your previous comment :

> This should be updated as well (based on the discussions I've seen in the pull request), since the asset code might be something outside of ISO4217.

I'm confused whether we want to encourage the use of ISO 4217 or keep it open to other conventions, and if we opt for the latter, then what are other possible codes that can be referenced besides cryptocurrencies?

hajjimo: I removed the mention of cryptocurrencies and left the ISO 4217 examples without stating that an ISO 4217 must be used for the assetCode.

PR #634, opened 2025-07-08
add diagram names

melissahenderson:
Added diagram names to sequence diagrams and removed client:load from wrapper as it's no longer necessary. Also made some minor text changes to appease Vale.


docs/src/content/docs/concepts/op-flow.mdx
```diff
@@ -18,14 +18,14 @@ This page provides a high-level look at the API calls and interactions that occu
 
 ### Assumptions
 
-- The <Tooltip content="Such as a mobile or web application or service" client:load>client's</Tooltip> user is the sender.
+- The <Tooltip content="Such as a mobile or web application or service">client's</Tooltip> user is the sender.
```
brad-dow, line 21: Just checking to make sure we want to remove `client:load` from the Tooltip component too

brad-dow (approved):

docs/src/content/docs/concepts/op-flow.mdx
melissahenderson: I'm not certain. I'm assuming we don't need it across the board. But if this breaks, then we know better 😬

PR #638, opened 2025-07-09
feat: add sepa payment method

sanducb:

- Add SEPA as payment method


mkurapov:
@sanducb was testing netlify preview 😅 

but I will close this for now now, as we do not need SEPA in OP

Issue #636, opened 2025-07-08
[DOCS] Use native code blocks instead of home-built component


melissahenderson:
When we first moved to Starlight, its native code block component didn't support titles. HJ created [our own component](https://interledger.net/shared/codeblock/) to support that. 

The native component [now supports titles](https://starlight.astro.build/guides/authoring-content/#frames-and-titles). We should start using the native component unless we have a good reason not to in a particular instance.

Update the following pages to use the native component.
* /concepts/payments
* /identity/http-signatures
* /identity/hash-verification

There may be other pages, but those are the ones I quickly identified.

melissahenderson:
Fixed by #640

PR #640, opened 2025-07-10
docs: mi/636/native-code-block

hajjimo:
Fixes issue # 636

- Replaced custom component with native code block support in static code examples

- Removed CodeBlock imports from doc files and references from Readme


melissahenderson:
Fixes #636

PR #646, opened 2025-07-28
docs: new set of guide updates

melissahenderson:
getting-started: removed links to guides for now. since the guides are becoming more specific, linking a "getting started" overview about p2p payments to a guide about making an online purchase could be confusing.

make-onetime-payment: updated content, removed snippets, changed file name to make-onetime-purchase to better reflect purpose of the guide

make-recurring-payments: updated content, removed snippets, changed file name to make-recurring-sub-payment to better reflect purpose of the guide

split-payments: removed unused card component import

code-snippets-template: deleted, never used


docs/src/content/docs/guides/make-onetime-purchase.mdx
```diff
@@ -0,0 +1,311 @@
...
+      access: [
+        {
+          identifier: customerWalletAddress.id,
+          type: 'outgoing-payment',
+          actions: ['read', 'create'],
+          limits: {
+            receiveAmount: {
+              assetCode: 'EUR',
+              assetScale: 2,
+              value: '3600'
```
brad-dow: Should this be 3500? I'm thinking about €35 throughout this guide, so I wasn't sure about the extra €1.

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
```diff
@@ -0,0 +1,383 @@
...
+The interval used in this guide is `R12/2025-05-20T13:00:00Z/P1M`. Remember that the customer wants to send payments of $20 USD a month for 12 months. The interval breaks down like this:
+
+- `R[12]/` is the number of repetitions - 12
+- `2025-05-20` is the start date of the repeating interval - 20 May 2025.
+- `T13:00:00Z/` is the start time of the repeating interval - 1:00 PM UTC.
+- `[P1M]` is the period between each interval - one month. Used with `R[12]`, you have a grant that's valid once a month for 12 months.
+
+Altogether, a $20 USD payment can be made from:
+
+- 1 PM UCT on 20 May 2025 through 12:59 PM UCT on 20 June 2025
```
brad-dow: UTC, not UCT

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
```diff
@@ -0,0 +1,383 @@
...
+- 1 PM UCT on 20 June 2025 through 12:59 PM UCT on 20 July 2025
```
brad-dow: UTC, not UCT

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
```diff
@@ -0,0 +1,383 @@
...
+    )
+    ```
+  </TabItem>
+</Tabs>
+
+### 12. Request the creation of another outgoing payment resource
+
+At the next valid interval, your app must request the creation of another outgoing payment resource on the customer's wallet account.
+
+Use the access token returned in Step 5, or the newly rotated access token returned in Step 11, to call the [Create Outgoing Payment API](/apis/resource-server/operations/create-outgoing-payment).
```
brad-dow: Similar to step 10, I'm not seeing an access token returned in Step 5. I think one is returned in Step 2 for the incoming payment, one in Step 4 for the quote resource, Step 6 for the outgoing payment, Step 9 to continue an outgoing payment, and Step 11 for rotating an access token where needed.

Should this one point back to Step 6 for the creation of a new outgoing payment grant?

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
```diff
@@ -0,0 +1,383 @@
...
+  </TabItem>
+</Tabs>
+
+### 10. Request the creation of an outgoing payment resource
+
+Recall that the Create Quote API response for the subscription service (Step 5) included a `debitAmount` and a `receiveAmount`. The response also included an `id` which is a URL to identify the quote.
+
+Because the quote contains debit and receive amounts, we won’t specify any other amounts when making a Create Outgoing Payment API request. Instead, we will specify a `quoteId`.
+
+Use the access token returned in Step 5 to call the [Create Outgoing Payment API](/apis/resource-server/operations/create-outgoing-payment).
```
brad-dow: I'm not seeing an access token returned in Step 5. I think one is returned in Step 2 for the incoming payment, one in Step 4 for the quote resource, Step 6 for the outgoing payment, Step 9 to continue an outgoing payment, and Step 11 for rotating an access token where needed.

Should this one point back to Step 9 for the token from the continuation?

docs/src/content/docs/guides/make-onetime-purchase.mdx
```diff
@@ -0,0 +1,311 @@
...
+    Create an Outgoing Payment
+  </LinkOut>
+
+## Steps
+
+### 1. Get wallet address information
+
+When a customer initiates a payment, you must get the wallet address information for the customer and for the store.
+
+Let's assume the store's wallet address is built into your app. Let's also assume the customer provided their wallet address at the beginning of the checkout flow.
```
hajjimo: It may confuse readers why the developer would fetch the wallet addresses if they already have them.  We could add text or an aside that mentions this step is necessary to obtain the `authServer` and `resourceServer` details on the accounts.

docs/src/content/docs/guides/make-onetime-purchase.mdx
```diff
@@ -0,0 +1,311 @@
...
+import StartInteraction from '/src/partials/grant-start-interaction.mdx'
+import FinishInteraction from '/src/partials/grant-finish-interaction.mdx'
+
+:::tip[Summary]
+Learn how to create a single, non-recurring payment with an `incomingAmount` addressed to a single recipient.
+:::
+
+Making a one-time payment (OTP) to a single recipient is one of the more common transaction flows. OTPs facilitate a number of use cases, such as online purchases, peer-to-peer payments, and remittances.
+
+Depending on the use case, the transaction will have a defined `incomingAmount`, `debitAmount`, or `receiveAmount`.
```
hajjimo: I'm fuzzy on which use cases require which objects.  Specifying the various use cases is out of scope for a guide on a specific use case, but perhaps mentioning which objects are necessary for this particular example might be useful.

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
```diff
@@ -0,0 +1,383 @@
...
+    Rotate an Access Token
+  </LinkOut>
+
+## Steps
+
+### 1. Get wallet address information
+
+When the customer initiates the payment for their subscription, you must get the wallet address information for the customer and the subscription service.
+
+Let’s assume the subscription service's wallet address is built into your app. Let’s also assume the customer provided their wallet address at the beginning of the checkout flow.
```
hajjimo: Same comment as in the one-time payment doc, if we assume the developer already has the wallet addresses, then why fetch them?  I assume the developer needs to get the `authServer` and `resourceServer` for the accounts, in which case we might mention it inline or in an aside.

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
```diff
@@ -0,0 +1,383 @@
...
+
+Depending on the use case, each transaction will have a defined `incomingAmount`, `debitAmount`, or `receiveAmount`.
+
+- `incomingAmount` - The maximum amount to pay toward an incoming payment. It’s possible to pay less, but not more.
+- `debitAmount` - A fixed amount to pay from the sender toward an incoming payment. It’s possible for the incoming payment to receive less than the `debitAmount`, depending on fees and other factors.
+- `receiveAmount` - A fixed amount to pay into an incoming payment. It’s possible for the amount paid by the sender to be greater than the `receiveAmount` depending on fees and other factors.
+
+In this guide, you'll assume the role of an app developer for a video-on-demand service. The guide takes you through the steps your app will perform to set up a recurring payment for a new subscriber. Your app will set an `incomingAmount` when creating the incoming payment resource and an `interval` in the outgoing payment grant. In this example, the subscriber will pay $20 USD a month for their selected plan.
+
+## Endpoints
```
hajjimo, line 23: I just noticed we LinkOut for API specs under Endpoints, but we link to the API specs on the page.   I wonder if we should LinkOut in all cases, so that a developer who needs to reference the API while going through the steps in a guide doesn't lose their place.

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
```diff
@@ -0,0 +1,383 @@
...
+            {
+              identifier: customerWalletAddress.id,
+              type: 'outgoing-payment',
+              actions: ['read', 'create'],
+              limits: {
+                interval: 'R12/2025-05-20T13:00:00Z/P1M'
+                debitAmount: {
+                  assetCode: 'USD',
+                  assetScale: 2,
+                  value: '2000',
```
hajjimo, line 230: In this example, the `debitAmount` doesn't have any additional fees as in the one-time purchase example.  It has me wondering if the use cases determine whether fees are added to `debitAmount`.  I imagine if ILP is used as the payment method, and the underlying ASEs are not direct peers on the Interledger network, then there probably will be a fee irrespective of the use case.

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
melissahenderson: Thank yooooou I must have hackathon on the brain

docs/src/content/docs/guides/make-onetime-purchase.mdx
melissahenderson: Good catch. Initially I had an example where a €1 EUR fee was added.

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
melissahenderson: Yes, Step 6 ty

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
melissahenderson: I'm going to do some revisions. Your comments made me realize something's not quite right. Why get an access token in the outgoing payment grant response AND the grant continuation response? I think I see where I messed up.

docs/src/content/docs/guides/make-onetime-purchase.mdx
melissahenderson: Good point. It might be distracting to have all of them listed. I'll think about this some but ill probably take the list out.

docs/src/content/docs/guides/make-recurring-sub-payment.mdx
melissahenderson: I included a fee in one of the other guides and Max suggested I remove it because it's a bit more complicated than just throwing a debit amount in somewhere. I can see us having some kind of fees guide in the future.

docs/src/content/docs/guides/make-onetime-purchase.mdx
melissahenderson: Fair. I have this repeated in most of the guides, I think. So I'll figure out a better way to explain.

melissahenderson:
@mkurapov No takers thus far in the open-payments channel. Please feel free to delegate.

PR #653, opened 2025-07-31
chore: chunked snippet experiment with snippets inside docs

raducristianpopa:
**_EXPERIMENT_** _based on the discussion that took place on 30th of July._

### DIFF
--- a/docs/src/components/ChunkedSnippetExperimental.astro
+++ b/docs/src/components/ChunkedSnippetExperimental.astro
@@ -1,0 +1,231 @@
+---
+import  fs  from "node:fs/promises"
+import path from "node:path"
+import { execSync } from 'node:child_process';
+import prettier from 'prettier';
+import { Code } from 'astro:components';
+import { CodeBlock } from '@interledger/docs-design-system'
+const {source, chunk, version, l} = Astro.props;
+import * as phpPlugin from '@prettier/plugin-php'
+
+const gitRoot = execSync('git rev-parse --show-toplevel').toString().trim();
+const snippetsPath = path.join(gitRoot, 'snippets');
+
+const getChunk = async () => {
+  const data = await fs.readFile(path.join(snippetsPath, version, l, source), {encoding: "utf8"})
+  return data
+}
+
+const code = await getChunk()
+
+// Split the text into an array of lines
+const lines = code.split('\n');
+
+// Initialize an array to store the delimiter objects
+const delimiterArray = [];
+
+// Iterate through the lines, extract "//@!" lines and format them into objects
+let startLine = null;
+for (let i = 0; i < lines.length; i++) {
+  if (lines[i].trim().startsWith("//@!")) {
+    if (startLine !== null) {
+      // Extract the entire "start_line" and the "title"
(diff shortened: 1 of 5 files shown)


docs/src/content/docs/1.0/version.ts
```diff
@@ -0,0 +1 @@
+export const VERSION = 'v1'
```
raducristianpopa, line 1: We can have a version in:
- src/docs/content/version.ts (for latest)
- src/docs/content/1.0/version.ts (for v1)

This version can be used with the new ChunkedSnippet component which will help to detect the correct snippet associated with the major version:

```mdx
<ChunkedSnippetExperimental source="token/token-revoke.ts" chunk="3" l="node" version={VERSION}  />
```

At the moment:
- `l`: language (i.e. node, rust, etc...) (NB: needs a better name, it was named `l` since `lang` and `language` were already used in the component)
- `source`: the file source relative to the root of the snippets folder
- `version`: by providing the version the component will read the correct version for the snippets

The path looks lile this: `<git-root>/snippets/${version}/${language}/${source}`

raducristianpopa: CC @mkurapov

docs/src/components/ChunkedSnippetExperimental.astro
```diff
@@ -0,0 +1,231 @@
...
+const {source, chunk, version, l} = Astro.props;
+import * as phpPlugin from '@prettier/plugin-php'
+
+const gitRoot = execSync('git rev-parse --show-toplevel').toString().trim();
+const snippetsPath = path.join(gitRoot, 'snippets');
+
+const getChunk = async () => {
+  const data = await fs.readFile(path.join(snippetsPath, version, l, source), {encoding: "utf8"})
+  return data
+}
```
raducristianpopa, line 17: This is the only code change that happened in this file - how we retrieve the source: `readFile` instead of a HTTP request.

docs/src/content/docs/1.0/version.ts
mkurapov: What do you think if we did version without the `v`, to allow for minor version + keep inline with the docs as well?
eg `<git-root>/snippets/1.0/node/${source}`

raducristianpopa: I am okay with both approaches. But, thinking more about it, would it be better to name the versions like `1.x` `2.x` if we only care about versioning major releases? This makes it clearer that we're referring to a major version, rather than a specific release.

raducristianpopa: Bringing more context:

Me and Max played with versioning a bit and we noticed that the OpenAPI definitions are not versioned by default. In the Astro config we will have to manually add the entries for `1.0/openapi/....`. 

One thing that we found out is that we will not be able to use dots for the path (`1.0/openapi/...`) and we will have to go for something like `v1/openapi/....`.

melissahenderson:
@mkurapov @raducristianpopa Is this OK to close?

mkurapov:
Closing as completed in https://github.com/interledger/open-payments/pull/689

PR #659, opened 2025-08-07
feat: add open-payments-specifications as a submodule 

mkurapov:
- Adding [open-payments-specifications](https://github.com/interledger/open-payments-specifications) as a submodule in the repo, pinning to [version 1.0.2
](https://github.com/interledger/open-payments-specifications/tree/v1.0.2)
- Update documentation to generate API docs from the specs in the submodule
- Remove workflows associated with the specs

**This effectively removes the [new `subject` field addition](https://github.com/interledger/open-payments/pull/590) from the API docs. This is desired, because the use of `subject` is undocumented, and not yet supported by SDKs/Rafiki.**

For now, openapi folder is not removed, because the node SDK uses it. It will be removed once we move everything to `open-payments-node` repository.

## Context

Fixes OPE-235


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "open-payments-specifications"]
```
raducristianpopa, line 1: Oh, I remember now... We cannot use tags for submodule, we have to checkout to a certain commit when we want to use a specific version 😢 .

I saw that it supports branching, maybe this is going to be cleaner, but it will require us to change how we do releases in the specification repository, or we can keep it the same and just create a new branch for the release? 🤔

mkurapov: I think this is OK for us, we shouldn't overwrite tags, so [eda1687](https://github.com/interledger/open-payments-specifications/tree/eda1687f0ba062a98ed311775d71255b7f1d2416) should always correspond to the [v1.0.2](https://github.com/interledger/open-payments-specifications/releases/tag/v1.0.2) tag.

If you mean by quickly being able to tell what version of the specs are being used, VERSION file should be easy enough.

PR #662, opened 2025-08-08
chore: remove Node SDK packages from repo

mkurapov:
- Remove all Node packages from repository, keep the docs folder only
- Update links/references in documentation to Node SDK to the new repository

## Context

Fixes OPE-236


docs/package.json
```diff
@@ -22,7 +22,6 @@
     "starlight-versions": "^0.5.4"
   },
   "devDependencies": {
-    "@prettier/plugin-php": "^0.24.0",
-    "prettier": "3.6.2"
```
mkurapov, line 25: we have prettier in the root repo

pnpm-workspace.yaml
```diff
@@ -1,3 +1,2 @@
 packages:
-  - 'packages/*'
   - 'docs'
```
mkurapov, line 2: For now, keeping this as a monorepo (and not moving everything from `docs/` to the root, since we will also have a `snippets/` directory, to keep things separated.

Suggestions welcome

raducristianpopa: I was thinking that we can move away from the monorepo setup for the docs. For the Node snippets, I would expect a separate package which is not part of the monorepo. This way we avoid polluting the root `node_modules` and we can install only the documentation dependencies. 🤔

mkurapov: @raducristianpopa that's fine as well. I was thinking we can standardize a few packages like prettier & typescript in the root, but I don't have a strong preference, esp. if you want to keep the runnable Node snippets standalone.

I can do that in a separate PR.

raducristianpopa (approved):

README.md
```diff
@@ -21,7 +21,11 @@ The Open Payments APIs are a collection of three sub-systems:
 - A **resource server** which exposes APIs for performing functions against the underlying accounts
 - A **authorisation server** which exposes APIs compliant with the [GNAP](https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol) standard for getting grants to access the resource server APIs
 
-This repository hosts the documentation for the Open Payments APIs, which is published on [openpayments.dev](https://openpayments.dev)
+The three Open Payments OpenAPI specifications are found in the [open-payments-specifications](https://github.com/interledger/open-payments-specifications) repository, while this repository hosts the documentation for the APIs, published on [openpayments.dev](https://openpayments.dev). 
+
+### Open Payments SDKs
+
+The Open Payments SDKs provide developers with pre-built functions that simplify interactions with the Open Payments API. Currently, we offer a [TypeScript/NodeJS SDK](https://github.com/interledger/open-payments-node/tree/main/packages/open-payments) and a [PHP SDK](https://github.com/interledger/open-payments-php), with plans to expand to additional languages and frameworks in the future.
```
raducristianpopa: We can mention the Rust one as well.

mkurapov: Added

raducristianpopa (approved):

PR #687, opened 2025-09-10
docs: recurring guide updates

melissahenderson:
Stopgap updates to recurring guide to help upcoming hackathon participants. More significant updates are coming in the next few weeks.

**Description of changes**

**Required**

- [X] Used LinkOut component on external links
- [ ] Reviewed Vale errors and made changes where appropriate - NA
- [X] Ran Prettier
- [X] Previewed updates in Netlify
- [ ] Received SME and/or peer approval if updates are significant
- [ ] Included a "fixes #" comment - NA

**Conditional**

- [ ] Ensured sequence diagrams follow our style guide - NA
- [X] Included code samples where appropriate
- [ ] Updated related READMEs - NA

### DIFF
--- a/docs/src/content/docs/guides/make-recurring-payments.mdx
+++ b/docs/src/content/docs/guides/make-recurring-payments.mdx
@@ -73,7 +73,7 @@ Outgoing payments require an interactive grant. This type of grant will obtain t
 Include the following in the request:
 
 - `limits` object
-  - `debitAmount` - The maximum amount that the user can send per interval (for example, per month). When the next interval begins, the value resets.
+  - `debitAmount` - The maximum amount that the user will pay per interval (for example, per month). When the next interval begins, the value resets.
   - `interval` - The [time interval](#about-the-interval) under which the grant is valid.
 
 Remember that your user, the sender, wants to make payments of $50 a month for three months. The amount resets each month and any unspent portions don't roll over.
@@ -93,7 +93,7 @@ Remember that your user, the sender, wants to make payments of $50 a month for t
             type: 'outgoing-payment',
             actions: ['read', 'create'],
             limits: {
-              interval: 'R3/2025-03-12T23:25:00Z/P1M'
+              interval: 'R3/2025-03-12T23:25:00Z/P1M',
               debitAmount: {
                 assetCode: 'USD',
                 assetScale: 2,


melissahenderson:
@mkurapov I wasn't sure what to do with rotate access token at the end, so I took the snippet out and referred to the API doc. More than happy to put a code example in there, just let me know what that should look like.

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,479 @@ title: Make recurring payments
...
+</Tabs>
 
-### 5. Create a Quote
+<details>
+  <summary>Example response</summary>
+  ```json wrap
+  {
+    "access_token": {
+      "value": "...", // access token value for incoming payment grant
+      "manage": "https://happylifebank.example.com/token/{...}", // management uri for access token
```
BlairCurrey: should be auth server url

```suggestion
      "manage": "https://auth.happylifebank.example.com/token/{...}", // management uri for access token
```

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,479 @@ title: Make recurring payments
...
+          "type": "incoming-payment",
+          "actions": ["create", "read"]
+        }
+      ]
+    },
+    "continue": {
+      "access_token": {
+        "value": "..." // access token for continuing the request
+      },
+      "uri": "https://happylifebank.example.com/continue/{...}" // continuation request uri
```
BlairCurrey: should be auth server url

```suggestion
      "uri": "https://auth.happylifebank.example.com/continue/{...}" // continuation request uri
```

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,479 @@ title: Make recurring payments
...
+      "assetCode": "USD",
+      "assetScale": 2
+    },
+    "receivedAmount": {
+      "value": "0",
+      "assetCode": "USD",
+      "assetScale": 2
+    },
+    "completed": false,
+    "createdAt": "2025-03-12T23:20:50.52Z",
```
BlairCurrey: I dont think i'll recommend any change here, but I just wanted to note that when I run this request against Rafiki, I get an `expiresAt` and `updatedAt`. Looks like we fall back to some default duration for expiration when not provided.

However, I checked the open payments spec and these fields arent required, which Rafiki is but one implementation of (in theory). Of course, these are open payment docs, not Rafiki, so that is the authority. 

Just mentioning because we could opt to omit other non-required fields if we want. For example the amounts arent required but are probably useful... but what about methods, since we are fuzzing it anways? 

I dont have a strong opinion here, just food for thought.

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,479 @@ title: Make recurring payments
...
+</Tabs>
 
-The `debitAmount` and `receiveAmount` parameters, which apply to the sender and recipient respectively, allow you to define the limits under which an outgoing payment can be created. Select the parameter most appropriate for your particular use case. Specify the `receiveAmount` when the recipient must receive a defined amount. Specify the `debitAmount` for cases when the sender can exercise a limit on the amount that can be debited from their account.
+<details>
+  <summary>Example response</summary>
+  ```json wrap
+  {
+    "access_token": {
+      "value": "...", // access token value for quote grant
+      "manage": "https:/cloudninebank.example.com/token/{...}", // management uri for access token
```
BlairCurrey: should be the auth server url (also missing a `/`)
```suggestion
      "manage": "https://auth.cloudninebank.example.com/token/{...}", // management uri for access token
```

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,479 @@ title: Make recurring payments
...
+        },
+      },
+    },
+  );
+  ```
+</TabItem>
+</Tabs>
 
-Add the authorization server’s interaction reference as the `interact_ref` value to the body of your continuation request.
+#### About the interval
```
BlairCurrey: What about putting the example response right after the request? I can see why we'd want to comment on the interval right after its used, but I feel like the request/response pair is still important to keep together, and normally the response should be minimized so it shouldnt make the interval/interval comment feel disjointed

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,479 @@ title: Make recurring payments
...
+  {
+    "access_token": {
+      "value": "...", // final access token required before creating an outgoing payment
+      "manage": "https://auth.cloudninebank.example.com/token/{...}", // management uri for access token
+      "access": [
+        {
+          "type": "outgoing-payment",
+          "actions": ["create", "read"],
+          "identifier": "https://cloudninebank.example.com/sender",
+          }
```
BlairCurrey: Rafiki returns the `limits` in the access array here although the spec doesnt require it. So I think it's your call based on if its helpful or not. In this case it would be the same as the given one.

```
          limits: {
            interval: 'R3/2025-03-12T23:25:00Z/P1M'
            debitAmount: {
              assetCode: 'USD',
              assetScale: 2,
              value: '5000',
            },
          },
```

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,479 @@ title: Make recurring payments
...
+      "value": "5000",
+      "assetCode": "USD",
+      "assetScale": 2
+    },
+    "sentAmount": {
+      "value": "0",
+      "assetCode": "USD",
+      "assetScale": 2
+    },
+    "createdAt": "2025-03-12T23:27:54.52Z"
```
BlairCurrey, line 378: Rafiki also returns updatedAt, failed, grantSpentReceiveAmount, grantSpentDebitAmount although they arent required by the spec and dont add much IMO

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,384 @@ title: Make recurring payments
...
+  }
+  ```
+</details>
 
-### 11. Rotate the Access Token
+The first of the three recurring payments is now set up. At the next interval (for example, one month from now), repeat Steps 6 - 8. You don't need to request a new outgoing payment grant (Step 4) because the original grant is still valid.
```
mkurapov: Likewise for the incoming payment grant (we would just need to rotate the token, for example)

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,384 @@ title: Make recurring payments
...
+      accessToken: senderOutgoingPaymentGrant.access_token.value
+    },
+    {
+      walletAddress: senderWalletAddress.id,
+      incomingPayment: 'https://happylifebank.example/com/incoming-payments/{...}',
+      debitAmount: {
+        value: '5000',
+        assetCode: 'USD',
+        assetScale: 2
+      },
```
mkurapov: If the incoming payment has an `incomingAmount`, its optional to include `debitAmount`.

docs/src/content/docs/guides/make-recurring-payments.mdx
melissahenderson: I was thinking that if it wasn't required, we could leave it out. But if Rafiki returns it (whether the person following this guide is using Rafiki or not would be unknown), I should probably put it back in.

docs/src/content/docs/guides/make-recurring-payments.mdx
melissahenderson: Hmm. The grantSpents are fairly new, I think. Maybe I will leave what you listed out of this iteration. The updates to this guide are kind of a stop gap. We want the guide to be usable for the upcoming hackathons, but the long term plan is to get rid of it completely and replace it with three other guides (recurring debitAmount, recurring receiveAmount, recurring incomingAmount). Maybe those properties would be appropriate for one or more of the new guides.

docs/src/content/docs/guides/make-recurring-payments.mdx
melissahenderson: I'll put a link to this comment on our meeting agenda so we can discuss

docs/src/content/docs/guides/make-recurring-payments.mdx
melissahenderson: Quote steps were removed from the guide.

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,385 @@ title: Make recurring payments
...
+    },
+    {
+      access_token: {
+        access: [
+          {
+            identifier: senderWalletAddress.id,
+            type: 'outgoing-payment',
+            actions: ['read', 'create'],
+            limits: {
+              interval: 'R3/2025-03-12T23:25:00Z/P1M'
```
brad-dow: I _think_ this line should have a comma at the end:

`  interval: 'R3/2025-03-12T23:25:00Z/P1M',
`

brad-dow (approved):
I think one minor tweak is all that is needed. This update is such a big step up from the previous version!

docs/src/content/docs/guides/make-recurring-payments.mdx
```diff
@@ -3,241 +3,385 @@ title: Make recurring payments
...
+:::note
+Outgoing payments require an interactive grant. This type of grant will obtain the user's consent before an outgoing payment is made against their wallet account. You can find more information in the [Open Payments flow](/concepts/op-flow/#outgoing-payment) and [identity providers](/identity/idp) pages.
+:::
 
-Then add the `interval` property to the `limits` object, which is the interval period conforming to the <LinkOut href="https://en.wikipedia.org/wiki/ISO_8601#Repeating_intervals">ISO8601 repeating interval string format.</LinkOut>
+Include the following in the request:
 
-:::note[Limits]
+- `limits` object
+  - `debitAmount` - The maximum amount that the user can send per interval (for example, per month). When the next interval begins, the value resets.
```
hajjimo: This sentence didn't make sense to me initially until I read the subsequent section.  It could be removed, or it could stay since clarification is just a few characters away.

melissahenderson: I dont know if it will helped but I changed "that the user can send" to "that the user will pay".

PR #742, opened 2025-12-11
docs: update amounts.mdx

### LINKED ISSUE #739: [DOCS] Better describe debit and receive amounts
Update the Amounts concept page to explain debit amount and receive amount.

When the definitions are solid, check all the guides for places where we've defined what these two amounts are and make sure they reflect the correct information.

Older comment from Max prior to the Summit:

> Technically, debitAmount is how much will be charged to the customer in their asset/currency.

We have areas where we use things like "wants to pay", which isn't accurate.

melissahenderson:
**Description of changes**

**Required**

- [NA] Used LinkOut component on external links
- [X] Reviewed Vale errors and made changes where appropriate
- [X] Ran Prettier
- [X] Previewed updates in Netlify
- [ ] Received SME and/or peer approval if updates are significant
- [X] Included a "fixes #" comment

fixes #739 

**Conditional**

- [NA] Ensured sequence diagrams follow our style guide
- [NA] Included code samples where appropriate
- [NA] Updated related READMEs

### DIFF
--- a/docs/src/content/docs/concepts/amounts.mdx
+++ b/docs/src/content/docs/concepts/amounts.mdx
@@ -21,7 +21,7 @@ Amounts represent monetary values and are fundamental to every Open Payments ope
 Before learning about amount components, you should understand the difference between debit and receive amounts.
 
 - Debit amount - The total amount that a sender will be charged, in their asset/currency, with the completion of an outgoing payment
-- Receive amount - The amount that was or will be paid into the recipient's account in their asset/currency
+- Receive amount - The amount that will be paid into the recipient's account in their asset/currency
 
 ## Values
 


docs/src/content/docs/concepts/amounts.mdx
```diff
@@ -16,20 +16,33 @@ Amounts in Open Payments define monetary values using asset codes and asset scal
 
 Amounts represent monetary values and are fundamental to every Open Payments operation. Whether you're creating an incoming payment, requesting a quote, or tracking payment progress, every amount consists of three key components: a numerical `value`, an `assetCode`, and an `assetScale`. Understanding these components is crucial for building Open Payments applications.
 
+## Debit and receive amounts
+
+Before learning about amount components, you should understand the difference between debit and receive amounts.
+
+- Debit amount - The total amount that a sender will be charged, in their asset/currency, with the completion of an outgoing payment
+- Receive amount - The amount that was or will be paid into the recipient's account in their asset/currency
```
mkurapov:
```suggestion
- Receive amount - The amount that will be paid into the recipient's account in their asset/currency
```
or we say "Receive(d) amount", if we want to keep "that was paid"

PR #743, opened 2025-12-14
update umami url

JonathanMatthey:
**Description of changes**

simply update the umami url to a faster one 

It is served by the same backend but uwa will go through a formal load balancer and cdn. Not only that it uses a real interleger dns.

### DIFF
--- a/snippets/node/grant/grant-continuation.js
+++ b/snippets/node/grant/grant-continuation.js
@@ -1,6 +1,7 @@
 import dotenv from "dotenv";
 import { join } from "path";
 import { fileURLToPath } from "url";
+import { isFinalizedGrant } from "@interledger/open-payments";
 
 dotenv.config({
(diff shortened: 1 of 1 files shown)


snippets/node/grant/grant-continuation.js
```diff
@@ -1,6 +1,7 @@
 import dotenv from "dotenv";
 import { join } from "path";
 import { fileURLToPath } from "url";
+import { isFinalizedGrant } from "@interledger/open-payments";
```
JoblersTune, line 4: Where does this change come from? I would expect the js files to be autogenerated from the ts files.

JonathanMatthey: i was getting an error that it was missing, and it is being called later in that file.  but yeah it's odd - not really part of this PR

JoblersTune (approved):

PR #777, opened 2026-04-16
docs: document directed identity for non-interactive grants

brad-dow:
Fixes #DOC-34

Updated the following pages to document directed identity:

- client-keys.mdx - key registry, client request body, grant requests flow, and sequence diagram intro
- grants.mdx - clarify that directed identity clients provide their public key directly and don't require a key registry
- auth.mdx - included directed identity as an option for incoming payment and quote grant requests

Also, bumps the `open-payments-specifications` submodule to pick up the latest spec changes, including directed identity

**Checklist**

- PR title is prefixed with `docs:`
- PR title or description includes a `fixes #`
- You've run `pnpm format` and `pnpm lint`
- You've reviewed Vale errors and made changes where appropriate

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,6 @@
 # Macs
 .DS_Store
 
-# Claude Code
-.claude
-
 # Logs
 logs
 *.log
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,9 @@
 # Macs
 .DS_Store
 
+# Claude Code
+.claude
+
 # Logs
 logs
 *.log
--- a/docs/src/content/docs/identity/client-keys.mdx
+++ b/docs/src/content/docs/identity/client-keys.mdx
@@ -9,7 +9,7 @@ import {
 } from '@interledger/docs-design-system'
 
 :::tip[Summary]
-Client keys are unique identifiers that allow clients to sign Open Payments requests. They allow a client's identity to be verified through a public key registry.
+Client keys are unique identifiers that allow clients to sign Open Payments requests and verify their identity.
 :::
 
 All client requests in Open Payments are signed using a unique key that identifies the client to authorization and resource servers. All requests, except for new grant requests, will carry an access token that's bound to the key.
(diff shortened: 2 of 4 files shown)


hajjimo (approved):

docs/src/content/docs/identity/client-keys.mdx
```diff
@@ -9,7 +9,7 @@ import {
 } from '@interledger/docs-design-system'
 
 :::tip[Summary]
-Client keys are unique identifiers that allow clients to sign Open Payments requests. They allow a client's identity to be verified through a public key registry.
+Client keys are unique identifiers that allow clients to sign Open Payments requests and verify their identity.
```
melissahenderson: Who is "their"? Clients or requests?

.gitignore
```diff
@@ -1,6 +1,9 @@
 # Macs
 .DS_Store
 
+# Claude Code
+.claude
```
melissahenderson: Was this a Prettier addition?

melissahenderson (commented):
Looks good overall. Do you think we should add directed identity to the glossary?

brad-dow:
> Looks good overall. Do you think we should add directed identity to the glossary?

I waffled a bit on whether to include it or not. I feel like the rest of the glossary terms are foundational to Open Payments. Things like ASE and infrastructure pieces like auth server, etc. Directed identity feels like a layer deeper than those. However, because it's a "thing with a name" now across multiple pages it's worth including? I could go either way.

docs/src/content/docs/identity/client-keys.mdx
brad-dow: Good catch...I'll add "the client's identity".

.gitignore
brad-dow: I installed claude code locally for some experimentation, and I'm pretty sure it was an artifact from that. I pulled that back out, so it shouldn't show up anywhere else.

PR #794, opened 2026-05-18
docs: ase-developer-journeys

hajjimo:
Fixes # [DOC-41](https://linear.app/interledger/issue/DOC-41/user-doc-investigative-op-ase-implementation-guide)

Third iteration in separating developer and ASE journeys.  New ASE specific content was generated by AI and will be reviewed once the information architecture is agreed upon by the writers.

**Checklist**

- PR title is prefixed with `docs:`
- PR title or description includes a `fixes #`
- You've run `pnpm format` and `pnpm lint`
- You've reviewed Vale errors and made changes where appropriate


docs/src/content/docs/identity/client-keys.mdx
mkurapov, line 1: On this page, we have this section:
> Clients that identify themselves with a [wallet address](https://deploy-preview-794--openpayments.netlify.app/concepts/wallet-addresses/) make their key registry publicly accessible via a jwks.json endpoint.

Technically, it is the ASEs responsibility to allow clients to register their public keys, and host them at the jwks.json endpoint.

docs/src/content/docs/overview/getting-started.mdx
mkurapov, line 1: Should we add a section to the Payments section quickly outlining the resources, or do you think that is sufficiently covered in For Developers > Concepts > Resources?

I guess this is the challenge of this PR: finding where to place certain things that are generic OP concepts 😅

mkurapov:
I like this the direction of this! Makes it easy to "choose your own experience".

docs/src/content/docs/identity/client-keys.mdx
hajjimo: Updated text in docs/identity/client-keys.mdx:

For clients that identify themselves with a wallet address, the ASE provides a way for clients to register public keys and hosts the client’s key registry at WALLET_ADDRESS/jwks.json. An authorization server can retrieve the client’s key registry from that endpoint.

Updated text in docs/implement/security.mdx:

 ASEs must provide a key registration flow for clients and host each client's JWKS at `WALLET_ADDRESS/jwks.json`.

docs/src/content/docs/overview/getting-started.mdx
hajjimo: Certainly, by no means is this a prescriptive take, but I've uploaded the following as a possibility:

For Developers > Concepts > Resources > 
- Payment resource types (lands on the start of the page)
- incoming-payment (anchor link to incoming-payment)
- quote (anchor link to quote)
- outgoing-payment (anchor link to outgoing-payment)

hajjimo:
> I like this the direction of this! Makes it easy to "choose your own experience".

Considering it's OP Learning Month, it would be great if there were a way to get feedback from the new, focused visitors, but it would be confusing for people to be directed to both the live site and this version.

docs/src/content/docs/identity/client-keys.mdx
mkurapov: > For clients that identify themselves with a wallet address

I think we should write as though this is the default, because providing a key directly won't be common, given it's only for specific purposes.

mkurapov:
> Considering it's OP Learning Month, it would be great if there were a way to get feedback from the new, focused visitors

AB testing for the docs? :)

mkurapov (approved):
Looks good, just one minor [comment](https://github.com/interledger/open-payments/pull/794/changes#r3347996757)

hajjimo:
> Looks good, just one minor [comment](https://github.com/interledger/open-payments/pull/794/changes#r3347996757)

Do you also want to take a look, or perhaps you already have, at the new ASE pages?
For ASEs > 
-Overview
-Wallet address architecture
-Resource server
-Authorization server
-Identity provider integration
-Security 

Also, do we want to merge these changes during OP LM?  If all looks good here, I can check in with Bibi before we merge so that we (I) don't screw up any metrics she may be tracking, given this is a major change to the navigation.

melissahenderson:
I'd like to take another look. We also need to think about broken links/redirects and Spanish translations but that's a tomorrow project.

docs/src/content/docs/implement/identity-provider.mdx
```diff
@@ -0,0 +1,65 @@
...
+
+The client verifies the hash to confirm that the redirect actually came from the authorization server, then calls the continuation URI to retrieve the access token. The exact hash construction is described in [Security](/implement/security/).
+
+## Reference implementation
+
+<LinkOut href="https://rafiki.dev">Rafiki</LinkOut> provides a reference
+<LinkOut href="https://rafiki.dev/integration/deployment/services/auth-service/">
+  authorization service
+</LinkOut>
+implementation that includes support for integration with an IdP.
```
mkurapov:
```suggestion
<LinkOut href="https://rafiki.dev">Rafiki</LinkOut> provides a reference
<LinkOut href="https://rafiki.dev/integration/deployment/services/auth-service/">
  authorization service
</LinkOut> implementation that includes support for integration with an IdP.
```
to fix

Image

docs/src/content/docs/implement/resource-server.mdx
```diff
@@ -0,0 +1,96 @@
...
+      "type": "ilp",
+      "ilpAddress": "g.ilp.iwuyge987y.98y08y",
+      "sharedSecret": "1c7eaXa4rd2fFOBl1iydvCT1tV5TbM3RW1WLCafu_JA"
+    }
+  ]
+```
+
+ASEs must generate a fresh `sharedSecret` per incoming payment and must not reuse secrets across resources. ILP is currently the only payment method that integrates with Open Payments. ASEs that intend to settle over a different rail must wait for that rail's payment method to be defined and integrated.
+
+## Application layer vs. settlement layer
```
mkurapov: I would avoid the use of "application layer", in comparison with the settlement layer. This section can just be called "Settlement".

docs/src/content/docs/implement/resource-server.mdx
```diff
@@ -0,0 +1,96 @@
...
+- Open Payments does not touch funds, hold balances, or execute transfers. The ASE is fully responsible for that.
+
+ASEs must keep the two layers in sync. As funds arrive at the recipient's ASE, the resource server must update the `receivedAmount` on the corresponding incoming payment so clients have an accurate view of payment progress.
+
+## Token validation
+
+Every request to the resource server carries an access token issued by the ASE's authorization server. Before serving the request, the resource server must validate that the token:
+
+- Is well-formed and not revoked.
+- Covers the requested action (`create`, `read`, `list`, `complete`) and resource type (`incoming-payment`, `quote`, `outgoing-payment`).
```
mkurapov, line 92: We should also include `limits`, if the action concerns outgoing payment creation

docs/src/content/docs/implement/resource-server.mdx
mkurapov: I don't necessarily know if its worth mentioning, but its possible that there might not be any settlement between ASEs involved if, for example, the sender and the receiver have accounts at the same ASE

docs/src/content/docs/overview/for-developers.mdx
```diff
@@ -0,0 +1,72 @@
...
+Building with Open Payments means writing a client that calls the Open Payments APIs to set up payments between Open Payments-enabled accounts. The accounts themselves, and the servers that expose them, are operated by account servicing entities (ASEs).
+:::
+
+This page is the starting point for developers building applications, services, or other software that consume the Open Payments APIs.
+
+## What you'll build
+
+A client in Open Payments is software that:
+
+- Discovers an Open Payments-enabled account from its [wallet address](/concepts/wallet-addresses/).
```
melissahenderson: I know what you're saying, but we should be clear what "it" is.

docs/src/content/docs/overview/for-developers.mdx
```diff
@@ -0,0 +1,72 @@
...
+- Requests [grants](/identity/grants/) from the account's authorization server.
+- Creates `incoming-payment`, `quote`, and `outgoing-payment` resources to set up a payment.
+- Signs every request with a key the authorization server can verify.
+- Handles the interactive consent flow for outgoing payments.
```
melissahenderson: Maybe "Handles interactive consent flows" since `incoming` and `quote` _could_ be interactive. Just a thought.

docs/src/content/docs/overview/for-developers.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+A client never moves money. The ASEs on each side of the payment are responsible for settlement.
+
+## What to read
+
+If you're new to Open Payments, follow this path:
```
melissahenderson: Maybe something like "check out these resources". IMO a path is laid out and sequential. But really, I think the reader could go top-bottom, left-right, or whatever they want.

docs/src/content/docs/overview/for-developers.mdx
```diff
@@ -0,0 +1,72 @@
...
+  <LinkCard
+    title="Guides"
+    href="/guides/accept-otp-online-purchase/"
+    description="End-to-end walkthroughs for common payment scenarios."
+  />
+</CardGrid>
+
+## What's intentionally not on the developer path
+
+Some Open Payments documentation describes the responsibilities of the ASE, not the client such as how the wallet address server is structured, how the authorization server processes grants, how the resource server hands payments off to settlement, how the identity provider is integrated. Those pages live under the [For ASEs](/implement/) section and are intended for the engineers deploying the server stack.
```
melissahenderson:
```suggestion
Some Open Payments documentation describes the responsibilities of the ASE, not the client. This includes how the wallet address server is structured, how the authorization server processes grants, how the resource server hands payments off to settlement, and how the identity provider is integrated. Those pages live under the [For ASEs](/implement/) section and are intended for the engineers deploying the server stack.
```

docs/src/content/docs/overview/for-ases.mdx
```diff
@@ -0,0 +1,71 @@
+---
+title: Implementing Open Payments
+---
+
+import { LinkCard, CardGrid } from '@astrojs/starlight/components'
+import { LinkOut } from '@interledger/docs-design-system'
+
+:::tip[Summary]
+Implementing Open Payments means standing up the servers that make a financial institution's accounts Open Payments-enabled. This page is the entry point for engineers at banks, digital wallet providers, and mobile money operators who are building that server stack.
```
melissahenderson: Semantics, I know. But maybe we should use financial service provider instead of institution. What do you think?

melissahenderson:
```suggestion
Implementing Open Payments means standing up the servers that make a financial service provider's accounts Open Payments-enabled. This page is the entry point for engineers at banks, digital wallets, mobile payment systems, and other providers who are building that server stack.
```

docs/src/content/docs/overview/for-ases.mdx
```diff
@@ -0,0 +1,71 @@
...
+This page is the starting point for account servicing entity (ASE) implementors.
+
+## What ASEs build
+
+To make accounts Open Payments-enabled, ASEs operate three servers and integrate with an identity provider:
+
+- A **wallet address server** that returns public information about each Open Payments-enabled account.
+- A **resource server** that hosts the `incoming-payment`, `quote`, and `outgoing-payment` APIs.
+- An **authorization server** that processes grant requests, issues access tokens, and coordinates user consent.
+- An **identity provider** integration that authenticates the resource owner and collects explicit consent for outgoing-payment grants.
```
melissahenderson:
```suggestion
- An **identity provider** integration that authenticates the resource owner and collects explicit consent for interactive grants.
```

docs/src/content/docs/overview/for-ases.mdx
```diff
@@ -0,0 +1,71 @@
...
+  <LinkCard
+    title="Security"
+    href="/implement/security/"
+    description="Signature validation, client key resolution, and interaction hash generation from the server side."
+  />
+</CardGrid>
+
+## The rest of the docs
+
+The Developer Concepts pages (under Overview → Concepts and Identity and access management) describe what client developers need to know to call the APIs. ASEs are welcome to read them. They describe the same protocol from the other side, but they are not the authoritative source for ASE implementation requirements. The For ASEs section is.
```
melissahenderson:
```suggestion
The Developer Concepts pages (under For Developers → Concepts) describe what client developers need to know to call the APIs. ASEs are welcome to read them. They describe the same protocol from the other side, but they are not the authoritative source for ASE implementation requirements.
```

melissahenderson: Suggest removing "and Identity and access management" since it's a subsection of Concepts. 

I removed the last sentence because it seemed a little odd for it to not have a link (because there isn't a place to link to, thanks Starlight). Alternatively, we could bold or italicize "For ASEs".

docs/src/content/docs/overview/for-ases.mdx
```diff
@@ -0,0 +1,71 @@
...
+
+## Reference implementation
+
+<LinkOut href="https://rafiki.dev">Rafiki</LinkOut> is an open-source
+implementation of an Open Payments-compatible server stack. The ASE
+implementation guide describes the standard- and protocol-level requirements;
```
melissahenderson: Should we refer to an ASE implementation guide like it's a standalone thing?

docs/src/content/docs/overview/for-ases.mdx
```diff
@@ -0,0 +1,71 @@
+---
+title: Implementing Open Payments
```
melissahenderson, line 2:
```suggestion
title: Implementing Open Payments
next: Developer concepts
```

melissahenderson: Just a thought on wording. Without a different `next` title, the navigation at the bottom of the page just says "Wallet addresses". Since this is an ASE page, it might be weird to send someone to the Developer's wallet concepts page w/ no context.

docs/src/content/docs/implement/ase-overview.mdx
```diff
@@ -0,0 +1,40 @@
+---
```
melissahenderson, line 1: Don't hate me, but I don't like this page being titled index.mdx 🫣

I'd rather it be something like ase-overview... just a little more specific. I know a change like this touches a lot of files.

docs/src/content/docs/implement/index.mdx
```diff
@@ -0,0 +1,40 @@
+---
+title: For ASEs
+---
+
+import { LinkOut } from '@interledger/docs-design-system'
+
+:::tip[Summary]
+These pages describe what an account servicing entity (ASE) must build to make its accounts Open Payments-enabled. The rest of the Open Payments documentation describes what client developers need to know to call the APIs; this guide describes what ASEs need to build to serve them.
```
melissahenderson:
```suggestion
These pages describe what an account servicing entity (ASE) must build to make its accounts Open Payments-enabled. The rest of the Open Payments documentation describes what client developers need to know to call the APIs. This guide describes what ASEs need to build to serve them.
```

docs/src/content/docs/implement/index.mdx
```diff
@@ -0,0 +1,40 @@
...
+:::
+
+## What is an ASE?
+
+An account servicing entity (ASE) provides and maintains payment accounts for senders and recipients, and is a regulated entity in the countries it operates. Examples of ASEs include banks, digital wallet providers, and mobile money providers.
+
+When an ASE adopts the Open Payments standard, the customers' financial accounts at that ASE become Open Payments-enabled. Clients (apps, services, and other software) can then call the Open Payments APIs against those accounts to retrieve account details and issue payment instructions, without needing a custom integration per ASE.
```
melissahenderson:
```suggestion
When an ASE adopts the Open Payments standard, the customers' financial accounts at that ASE become Open Payments-enabled. Clients (apps, services, and other software) can then call the Open Payments APIs against those accounts to retrieve account details and issue payment instructions without needing a custom integration per ASE.
```

docs/src/content/docs/implement/index.mdx
```diff
@@ -0,0 +1,40 @@
+---
+title: For ASEs
```
melissahenderson: This page has For ASEs as the title at the top, then Overview as the title in the sidebar.

This means the bottom navigation for the page directly preceding it in the sidebar will read "Overview". 

The order of pages might change, so the preceding page might not always be "Get an outgoing payment grant for future payments". Using that guide as the example though, a reader would get to the bottom and maybe think "Overview of what?"

I don't think there's an elegant/tricky way to deal with this. I think the page name in the sidebar needs to change.

docs/src/content/docs/implement/resource-server.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+### quote
+
+When the resource server receives a <LinkOut href='/apis/resource-server/operations/create-quote'>Create a Quote</LinkOut> request, it must:
+
+- Compute the `debitAmount`, `receiveAmount`, and fee for the requested payment, including any applicable exchange rate when the sender's and recipient's wallet addresses use different `assetCode` values.
+- Commit to deliver the quoted `receiveAmount` to the recipient's ASE if an outgoing payment is created against the quote within the quote's validity window.
+- Assign the quote a `quoteId` and persist it.
+- Set a validity window after which the quote is no longer honored. ASEs define this window; it must be short enough that rate and fee exposure is bounded and long enough that the client can collect interactive consent before it expires.
```
melissahenderson: Is there a way to word this so that it doesn't sound like the interactive consent is collected on the create quote call?

docs/src/content/docs/implement/resource-server.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+A quote represents a binding commitment from the sender's ASE. If the sender's ASE cannot honor the quote at execution time, that is an ASE-side failure, not a protocol error.
+
+### outgoing-payment
+
+When the resource server receives a <LinkOut href='/apis/resource-server/operations/create-outgoing-payment'>Create Outgoing Payment</LinkOut> request, it must:
+
+- Validate that the access token presented carries a grant that authorizes this specific outgoing payment, including the limits collected during the interactive consent flow.
+- If a `quoteId` is present, look up the quote, confirm it has not expired, and use the quoted amounts.
+- If no `quoteId` is present (for example, in Web Monetization-style use cases), use the `incomingPayment` and `debitAmount` from the request to scope the outgoing payment.
```
melissahenderson: Maybe link to webmonetization.org? Not sure if we should try to direct their attention outward.

docs/src/content/docs/implement/resource-server.mdx
```diff
@@ -0,0 +1,96 @@
...
+- Persist the outgoing payment resource and initiate settlement.
+
+The outgoing-payment resource is the ASE's instruction to itself to move funds. Returning a `201` does not mean money has moved; it means the resource server has accepted and durably stored the instruction.
```
melissahenderson: So if we write outgoing payment resource, that's fine. But if we use outgoing-payment resource (w/ a hyphen), then `outgoing-payment` should be in the code style, since that's the technical name for it.

docs/src/content/docs/implement/resource-server.mdx
```diff
@@ -0,0 +1,96 @@
...
+"methods": [
+    {
+      "type": "ilp",
+      "ilpAddress": "g.ilp.iwuyge987y.98y08y",
+      "sharedSecret": "1c7eaXa4rd2fFOBl1iydvCT1tV5TbM3RW1WLCafu_JA"
+    }
+  ]
+```
+
+ASEs must generate a fresh `sharedSecret` per incoming payment and must not reuse secrets across resources. ILP is currently the only payment method that integrates with Open Payments. ASEs that intend to settle over a different rail must wait for that rail's payment method to be defined and integrated.
```
melissahenderson: Defined and integrated where? Maybe something like "must wait for that rail's payment method to be defined and integrated within the Open Payments specification". Or maybe standard? @mkurapov what do you think?

docs/src/content/docs/implement/resource-server.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+## Application layer vs. settlement layer
+
+Open Payments is a protocol that operates in the application layer. The resource server stores the agreement between the two ASEs ("ASE A will deliver X to ASE B against this incoming payment") and exposes it through the APIs.
+
+Actual movement of funds happens on the settlement layer:
+
+- Once the outgoing payment is created and the corresponding incoming payment is open, the sender's ASE is obligated to settle with the recipient's ASE.
+- Settlement runs over a shared payment rail outside of Open Payments.
```
melissahenderson: And outside of ILP, which might be worth mentioning since it's the only supported payment method atm.

docs/src/content/docs/implement/security.mdx
```diff
@@ -0,0 +1,99 @@
...
+2. Reconstruct the signature base from those components exactly as the client did.
+3. Resolve the public key (refer to [Client key resolution](#client-key-resolution) below).
+4. Verify the `Signature` header against the reconstructed signature base using Ed25519.
+5. Reject the request if any of the following are true:
+   - Required components (method, target URI, content digest where applicable, authorization header where applicable) are missing.
+   - The signature does not validate.
+   - The signature is older than the maximum age the ASE permits.
+   - The `Content-Digest` header is present but does not match the body.
+
+ASEs set their own maximum signature age. Open Payments does not bound it, but very long limits increase replay risk and very short limits can cause failures for clients on slow networks.
```
melissahenderson: What is a replay risk?

docs/src/content/docs/implement/security.mdx
```diff
@@ -0,0 +1,99 @@
...
+2. Issues a `GET` to `{walletAddress}/jwks.json` to retrieve the client's key registry.
+3. Locates the key whose `kid` matches the `keyId` in the `Signature-Input` header.
+4. Confirms the key uses the expected algorithm (`EdDSA`), key type (`OKP`), and curve (`Ed25519`).
+5. Binds the resolved key to the grant. On subsequent requests against the same grant, the authorization server fetches the bound key from the grant's stored domain rather than the request body.
+
+ASEs are responsible for keeping the hosted JWKS in sync with client key lifecycle events such as key rotation and key revocation.
+
+### Directed-identity clients
+
+For non-interactive grants (incoming-payment, quote), the client may provide its key directly in the request body as a `jwk`. The authorization server uses that key for signature verification on the initial request and binds it to the resulting grant. No JWKS endpoint lookup is required, and the client's wallet address is never exposed.
```
melissahenderson:
```suggestion
For non-interactive grants (`incoming-payment`, `quote`), the client may provide its key directly in the request body as a `jwk`. The authorization server uses that key for signature verification on the initial request and binds it to the resulting grant. No JWKS endpoint lookup is required, and the client's wallet address is never exposed.
```

docs/src/content/docs/concepts/resources.mdx
```diff
@@ -7,6 +7,10 @@ tableOfContents:
 
 import { LinkOut, Tooltip } from '@interledger/docs-design-system'
 
+:::note[For client developers]
+This page describes Open Payments from the client developer's perspective. ASEs should refer to the [For ASEs](/implement/) section.
+:::
+
 The Open Payments APIs are served by a resource server. Operations on the APIs require the client to have a valid access token issued by a trusted authorization server.
```
melissahenderson, line 18: This page needs a summary

docs/src/content/docs/concepts/payments.mdx
```diff
@@ -5,6 +5,10 @@ parent: Concepts
 
 import { LinkOut } from '@interledger/docs-design-system'
 
+:::note[For client developers]
+This page describes Open Payments from the client developer's perspective. ASEs should refer to the [For ASEs](/implement/) section.
+:::
+
 The payment method is the means by which the sender's ASE will fulfill its payment obligations to the recipient's ASE. Cash, credit/debit cards, bank transfers, gift cards and mobile money can all be considered different payment methods.
```
melissahenderson, line 12: This page needs a summary

docs/src/content/docs/concepts/resources.mdx
```diff
@@ -7,6 +7,10 @@ tableOfContents:
```
melissahenderson, line 10: I like what you've done to the sidenav w/ this page (including links for incoming-payment, quote, outgoing-payment). Unfortunately, it gets the bottom navigation stuck in a loop. The Next link points to incoming-payment, so you can't get to the Authorization page.

I think the solution is to add this to the frontmatter:

`next:`
 `  link: /concepts/auth`
`  label: Authorization`

You might have to play around w/ the indents. 

The only negative to this is if we change the order of the pages, we have to remember to change the frontmatter.

melissahenderson (commented):
I have a few comments. Some might be easier to go over in our team meeting on Monday. 

* For anything that's repeated across pages, we should make it a partial whenever possible. For ex: the Rafiki note at the bottom of some pages, the For client developers aside.
* However...I don't like having two stacked asides as the first thing you see when going through the developer concept pages. I haven't had a chance to think about alternatives.
* I'm also not sure how helpful it is to, for example, be on the developer Amounts page and have a link to the For ASEs overview page. That overview page doesn't have any information about amounts.

docs/src/content/docs/implement/resource-server.mdx
mkurapov: > ASEs that intend to settle over a different rail

Technically, these payment methods are not settlement layers/rails but more so messaging rails.

mkurapov: @melissahenderson yes, we should add something like "Open Payments standard" at the end of the sentence, if we want to keep it. Maybe we can instead turn this into a call-to-action, where ASEs can instead propose a new method to be added?

docs/src/content/docs/implement/identity-provider.mdx
hajjimo: f*n prettier

docs/src/content/docs/implement/resource-server.mdx
hajjimo: ## Settlement

Open Payments carries payment instructions; it does not move funds. The resource server records the agreement between parties ("ASE A will deliver X to ASE B against this incoming payment") and exposes it through the APIs.

The ASE is responsible for executing the payment against the underlying accounts:

- Once the outgoing payment is created and the corresponding incoming payment is open, the sender's ASE must deliver the agreed amount to the recipient.
- When the sender and recipient use different ASEs, settlement between those ASEs runs over a shared payment rail outside of Open Payments and ILP.
- When both parties hold accounts at the same ASE, the ASE may complete the payment by moving funds between those internal accounts. No inter-ASE settlement is required.
- Open Payments does not touch funds, hold balances, or execute transfers.

docs/src/content/docs/implement/resource-server.mdx
hajjimo: Every request to the resource server carries an access token issued by the ASE's authorization server. Before serving the request, the resource server must validate that the token:

- Is well-formed and not revoked.
- Covers the requested action (`create`, `read`, `list`, `complete`) and resource type (`incoming-payment`, `quote`, `outgoing-payment`).
- Is scoped to the wallet address the request targets, where applicable.
- Enforces grant `limits` on outgoing-payment `create` requests, including debit amount, receive amount, and interval restrictions collected during interactive consent.

docs/src/content/docs/overview/for-ases.mdx
hajjimo: Good catch! Thanks.

docs/src/content/docs/implement/index.mdx
hajjimo: I removed references to an "ASE guide" in this batch of updates.

docs/src/content/docs/implement/resource-server.mdx
hajjimo: ASEs must generate a fresh `sharedSecret` per incoming payment and must not reuse secrets across resources.

ILP is currently the only payment method defined in the Open Payments standard. Payment methods are messaging mechanisms, not settlement layers. ASEs that need a different payment method must wait for it to be defined and integrated in the standard, or propose adding one. Refer to [Get involved](/resources/get-involved/).

docs/src/content/docs/implement/security.mdx
hajjimo: ASEs set their own maximum signature age. Open Payments does not bind it. If the limit is too long, an intercepted signed request can be resent and still accepted (a replay attack). If the limit is too short, legitimate clients on slow networks may see their signatures expire before the request arrives.

docs/src/content/docs/concepts/resources.mdx
hajjimo: Ah, I failed to notice that. The frontmatter solution worked. Thanks!

docs/src/content/docs/overview/for-ases.mdx
hajjimo: Good point!  We should steer away from references to a "guide," since we ended up with a collection of pages with no required step sequence.

docs/src/content/docs/implement/index.mdx
hajjimo: Renamed to ASE overview.  I know it's not great, but I'm open to other options.

docs/src/content/docs/implement/resource-server.mdx
hajjimo: - Set a validity window after which the quote is no longer honored. ASEs define this window. It must be short enough to bind rate and fee exposure, and long enough for the client to complete any remaining grant and authorization steps and call Create Outgoing Payment before the quote expires.

docs/src/content/docs/implement/resource-server.mdx
melissahenderson: @mkurapov Are limits not optional? The Grant Request API spec doesn't list it as required.

docs/src/content/docs/implement/resource-server.mdx
melissahenderson: Only suggestion I have here is to change the ending punctuation for the list's introduction sentence from a colon to a period.

docs/src/content/docs/implement/resource-server.mdx
mkurapov: @melissahenderson limits are indeed required for outgoing-payment create requests

docs/src/content/docs/implement/resource-server.mdx
```diff
@@ -0,0 +1,98 @@
...
+## Token validation
+
+Every request to the resource server carries an access token issued by the ASE's authorization server. Before serving the request, the resource server must validate that the token:
+
+- Is well-formed and not revoked.
+- Covers the requested action (`create`, `read`, `list`, `complete`) and resource type (`incoming-payment`, `quote`, `outgoing-payment`).
+- Is scoped to the wallet address the request targets, where applicable.
+- Enforces grant `limits` on outgoing-payment `create` requests, including debit amount, receive amount, and interval restrictions collected during interactive consent.
+
+ASEs decide whether the resource server validates tokens locally (for example, against a shared cache) or by calling the authorization server. The protocol does not mandate one approach. Whichever pattern an ASE chooses, the resource server must refuse requests with insufficient tokens. For more details on token issuance and lifecycle, refer to [Authorization server](/implement/auth-server/).
```
mkurapov: Re:

> ASEs decide whether the resource server validates tokens locally (for example, against a shared cache) or by calling the authorization server. The protocol does not mandate one approach.

We can just be brief and say that ASEs can decide how and where the tokens are validated, as long as they are, instead of mentioning anything re: cache or services calling each other

docs/src/content/docs/implement/auth-server.mdx
```diff
@@ -0,0 +1,90 @@
...
+
+## Grant request processing
+
+Every grant request must be signed by the client. The authorization server's first job is to verify the signature. Refer to [Security](/implement/security/) for the verification rules.
+
+After signature verification, the authorization server inspects the request to determine the grant type and whether interaction is required.
+
+### incoming-payment grants
+
+Non-interactive. The authorization server may issue an access token immediately if the request is well-formed and the client is trusted.
```
mkurapov: Incoming payment grants may be interactive if the `list-all` action is requested, since `list-all` allows the client to paginate across all of the wallet address's incoming payments

docs/src/content/docs/implement/resource-server.mdx
melissahenderson: @hajjimo, I think what you have here is good.

docs/src/content/docs/implement/identity-provider.mdx
```diff
@@ -0,0 +1,59 @@
...
+## What the IdP must display
+
+The IdP is the only system in the flow that the resource owner is authenticated against, so it is also the only system that can credibly ask for consent. The IdP must show:
+
+- Who the client is (display name; optionally additional client attestation).
+- What the client is asking permission to do, in plain language (for example, "send up to $50 USD from your account").
+- Any time or velocity limits attached to the request.
+- A clear accept/deny choice.
+
+ASEs decide the exact UI, but the consent prompt must accurately reflect what the authorization server will encode in the resulting grant. If the IdP under-represents the requested access (for example, by omitting a `interval`), the grant must be denied or the access scaled down to what the user actually saw and accepted.
```
mkurapov: I think mentioning "the grant must be denied or the access scaled down" might be more confusing to the reader. I think we could just keep the first sentence, and then say something about how if the IDP does not accurately reflect the request received at the ASE, it exposes the ASE/user to risks (ie the user accepting a grant for more access that what was actually requested)

docs/src/content/docs/implement/auth-server.mdx
melissahenderson: @hajjimo In general, I don't recall if we discuss how to make non-interactive grants interactive. Can you do a quick search for that?

docs/src/content/docs/implement/security.mdx
```diff
@@ -0,0 +1,99 @@
...
+## HTTP signature validation
+
+Every request a client sends to an ASE's authorization or resource server is signed using the <LinkOut href='https://datatracker.ietf.org/doc/html/draft-ietf-gnap-core-protocol#name-http-message-signatures'>HTTP message signatures</LinkOut> profile defined by GNAP, with the Ed25519 variant of EdDSA.
+
+For every signed incoming request, ASEs must:
+
+1. Read the `Signature-Input` header to identify the `keyId` and the list of covered components.
+2. Reconstruct the signature base from those components exactly as the client did.
+3. Resolve the public key (refer to [Client key resolution](#client-key-resolution) below).
+4. Verify the `Signature` header against the reconstructed signature base using Ed25519.
```
mkurapov: Should mention the public key is necessary to verify `signature` or is it clear enough?

docs/src/content/docs/identity/client-keys.mdx
```diff
@@ -18,14 +19,16 @@ All client requests in Open Payments are signed using a unique key that identifi
 
 A key registry is a list of keys that are generated and stored by the client for when the client requires access to protected Open Payment resources. Grant requests are completed over multiple signed HTTP requests. As such, it's important that the client provides a way to consistently identify itself to an authorization server. The key registry allows the authorization server to verify that the client is who it says it is.
```
mkurapov: Technically the key registry (the public key registry) is hosted by the ASE's Open Payments resource server. How the client/developer interacts with the ASE to generate and host those keys is out of scope. The client simply needs to use the private key and the corresponding keyId when making requests such that they could be properly verified by the AS/RS that the client calls.

docs/src/content/docs/implement/auth-server.mdx
mkurapov: > how to make non-interactive grants interactive

Technically, interactive vs non-interactive grants are at the discretion at the ASE implementing Open Payments, but we can strongly urge to make outgoing payments & incoming payments list-all interactive, as we do in Rafiki

hajjimo: Added the following text to the auth page under Concepts, since we did not mention how non-interactive grants are made interactive in the OP docs.  

Incoming payment grants are non-interactive by default, meaning interaction by an individual (typically the client’s user) is not required for the authorization server to issue an access token. If the grant request includes the list-all action, the authorization server should require interaction before issuing an access token, since that action lets the client list incoming payments it did not create.

Quote grants are non-interactive by default, meaning interaction by an individual (typically the client’s user) is not required for the authorization server to issue an access token. If the grant request includes the list-all action, the authorization server should require interaction before issuing an access token, since that action lets the client list quotes it did not create.

docs/src/content/docs/implement/resource-server.mdx
mkurapov: @melissahenderson @hajjimo I wanted to correct myself here: limits are indeed optional for create outgoing payment grant requests as you pointed out @melissahenderson, but I believe it should not change the content on this page

PR #828, opened 2026-09-04
Demo/docs freshness

brad-dow:
This is a companion to the [docs freshness one-pager](https://docs.google.com/document/d/131bbD79QZfrZglTirPHO1qHiak92uKjTW6xyGCFhasE/edit?tab=t.0#heading=h.pedcxuo6gqyt) I previously shared. The goal here is to show the three areas we discussed in action (or as close to action as possible) so we have something before doing any actual implementation.

### Last reviewed date
I enabled the `lastUpdated` setting in `astro.config.mjs`. A few things to consider:

- Dates appear in the page footer by default (Starlight's built-in behavior)
- There's also a new custom component (`PageTitle`) that presents the last updated date at the top of the page
- Spanish pages localize automatically, which is nice (I didn't know how that would work)
- API reference pages don't show a date because they are generated from OpenAPI specs using the `starlight-openapi` plugin 

We need to decide if we want this date in the footer only, top only, or both. I'm assuming it's easier to compare dates if they are at the top of the page, but if we think the footer is acceptable, then I can remove the custom component altogether. **Update: We went with top only**

### Tracking translation drift
There aren't any code changes here, as `lastUpdated` kind of takes care of what we needed to compare page versions easier. With dates visible on both versions, we can quickly check from the docs instead of getting deep into GitHub.

I will say, Open Payments docs look pretty good right now because of the last batch of translations being updated. The only real exception I found was the verify wallet address ownership page, but that's because it's a new English guide without a translated version yet.

### Review automation for stale pages
I added the GitHub Actions workflow and script files on this branch. Here's what the output looks like when I ran the script locally against the repo with a 180-day threshold setting:

| Page | Last updated | Days since update  |
|-------|---------------|-----------------------|
| `resources/payment-methods.mdx` | 2024-07-29 | 767 days |
| `resources/glossary.mdx` | 2025-10-09 | 330 days |
| `identity/hash-verification.mdx` | 2025-11-18 | 290 days |
| `resources/further-learning.mdx` | 2025-12-03 | 275 days |
| `concepts/amounts.mdx` | 2025-12-16 | 262 days |

Once the workflow is tested and merged into main, it would run automatically on the 1st of every month and produce this same output.

`payment-methods.mdx` is a fossil at this point...that one is probably the most obvious one to look at first. `index.mdx` is interesting because it's the main landing page. It probably doesn't need constant upkeep, but that's why we are doing this exercise!

The workflow output lives in the GitHub Actions tab. Your idea of a monthly recurring Linear issue (the same way we do translations) is the best approach for now. A new Linear issue created will remind us to check the GitHub Actions page. We'll review the output there and then create any issues as needed. No need for automations between GitHub and Linear just yet.

### DIFF
--- a/.github/scripts/check-stale-docs.sh
+++ b/.github/scripts/check-stale-docs.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+DOCS_DIR="docs/src/content/docs"
+THRESHOLD_DAYS=180
+THRESHOLD=$(date -d "$THRESHOLD_DAYS days ago" +%s)
+
+echo "## Stale pages (not updated in $THRESHOLD_DAYS+ days)"
+echo ""
+
+found=0
+while IFS= read -r -d '' file; do
+  [[ "$file" == *"/es/"* ]] && continue
+  [[ "$file" == *"404"* ]] && continue
+
+  last_commit=$(git log -1 --format="%ct" -- "$file")
+  [[ -z "$last_commit" ]] && continue
+
+  if [[ "$last_commit" -lt "$THRESHOLD" ]]; then
+    last_date=$(git log -1 --format="%ci" -- "$file" | cut -c1-10)
+    days=$(( ($(date +%s) - last_commit) / 86400 ))
+    rel="${file#$DOCS_DIR/}"
+    echo "- $rel | Last updated: $last_date | ${days} days ago"
+    found=1
+  fi
+done < <(find "$DOCS_DIR" \( -name "*.mdx" -o -name "*.md" \) -print0)
+
+[[ $found -eq 0 ]] && echo "No stale pages found."
--- a/.github/workflows/stale-docs.yml
+++ b/.github/workflows/stale-docs.yml
@@ -1,0 +1,18 @@
+name: Stale docs check
+
+on:
+  schedule:
+    - cron: '0 9 1 * *'  # 9am on the 1st of every month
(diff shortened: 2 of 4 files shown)


melissahenderson:
1. Is there a way to exclude the landing page from showing the last updated date? 
2. I like the date at the top only. We can figure out how to handle "but what changed?" questions as they come up.
3. I don’t necessarily like using slashes, but what do you think about “Last updated/reviewed”? Which leads me to the next item.
4. Let’s assume payment-methods.mdx is fine. In this case, would we do something to force a change on the page (eg change a word) and push?

melissahenderson:
I've created a bare-bones recurring issue in Linear that will run on the 1st of the month. We can add details to the description as we figure it all out.

brad-dow:
> 1. Is there a way to exclude the landing page from showing the last updated date?
> 2. I like the date at the top only. We can figure out how to handle "but what changed?" questions as they come up.
> 3. I don’t necessarily like using slashes, but what do you think about “Last updated/reviewed”? Which leads me to the next item.
> 4. Let’s assume payment-methods.mdx is fine. In this case, would we do something to force a change on the page (eg change a word) and push?

1. Landing pages now exclude the dates
2. Date at top only
3. I went with "last reviewed" for now, but open to additional discussion here
4. We will use the frontmatter `lastUpdated` instead of touching other content

melissahenderson (approved):

